PR #4, opened 2017-06-29
MERGE: Add AICc

garyForeman:
Creating PR to let you know what I'm up to. This shouldn't pass Travis because all I've written so far are the test cases, not the working prod code.


tgsmith61591:
Make sure you also make `'aicc'` a valid optimization criterion in the `auto.py` file.

tgsmith61591:
Two things:
1. Can you change the PR target to `develop` branch rather than `master`?
2. I've added an [`if_has_delegate`](https://github.com/tgsmith61591/pyramid/blob/develop/pyramid/utils/metaestimators.py#L60) method in 4bb4800, which behaves similarly to the `if_delegate_has_method` call, and achieves what you were hoping to accomplish in the AICc WIP you showed me today. Once you merge, we can add that decorator in.

garyForeman:
Cool, will do. Looks like we'll have to think through the SARIMAX implementation a bit, since it doesn't have some of the needed attributes. I'll plan to make another commit tomorrow, so you can get a sense of current status and issues.

garyForeman:
I never leveraged your new decorator `if_has_delegate`. Let me know if you want me to do something with that.

tgsmith61591:
These changes look great, Gary. No worries on the decorator. I'll merge in and add the decorator. Thanks for your work on this!!

tgsmith61591:
@garyForeman this was merged into master for release 0.4. Thanks again for your help on this!

PR #8, opened 2017-07-02
Fix Setup.py for pypi

charlesdrotar:
This PR addresses minor inconsistencies that would have inhibited building the package and deploying it to pypi as a wheel.

This is necessary to address [issue#7](https://github.com/tgsmith61591/pyramid/issues/7).


tgsmith61591:
Does this pass registration on pypi test?

PR #11, opened 2017-07-04
Add Continuous Deployment to .travis.yml

charlesdrotar:
This PR adds continuous deployment to Pypi on tagged commits.

## Notable mentionings

- It will be important to tag commits before pushing so that Travis will build the wheels properly. I believe there is a means to tag it through github itself, which I am fairly certain will not trigger the build of the wheels.
- The wheels that are built do not have OS, architecture, and OS version tied to them. Essentially two wheels are built - one for python 2 and one for python 3. 

@tgsmith61591 Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,12 @@
 sudo: required
 language: python
-
 cache:
   apt: true
   directories:
   - $HOME/.cache/pip
   - $HOME/.ccache
-
 dist: trusty
-
 before_install: source build_tools/travis/before_install.sh
-
 env:
   global:
     # Directory where tests are run from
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -57,9 +53,21 @@ script: bash build_tools/travis/test_script.sh
...
+    on_failure: always
+    on_pull_requests: true
+
+deploy:
+  #server: https://test.pypi.org/legacy/ # Upload to pypi test server
+  distributions: "bdist_wheel" # Upload tar.gz and .whl files.
+  skip_upload_docs: true
+  skip_cleanup: true
+  on:
+    tags: true # Upload on tagged releases
```
tgsmith61591: What about on branch `master`? Or is the `tags` technically proxy for that, since we'd never tag a non `master` release?

setup.py
```diff
@@ -60,7 +60,8 @@
     print('Adding extra setuptools args')
     extra_setuptools_args = dict(
         zip_safe=False,  # the package can run out of an .egg file
-        include_package_data=True,
+        include_package_data=False,
```
tgsmith61591: Is there a particular reason for this? If we don't include package data, the `MANIFEST` (built from `sdist`) won't place the `.pyx` & `.c` files. But I guess we don't need those distributed... just for development. So maybe I answered my own question.

tgsmith61591: Actually, disregard this. I just looked through your commit history on your fork and answered my own question.

.travis.yml
charlesdrotar: Essentially 👍

PR #17, opened 2017-10-31
Add Linux wheel build with UCS-4 encoding for CPython 2.7

shoelsch:
I'm unable to install Pyramid from a Wheel on an Amazon Linux EC2 machine. Based on [PEP 513](https://www.python.org/dev/peps/pep-0513/#ucs-2-vs-ucs-4-builds), a `--enable-unicode` flag needs to be passed when building CPython 2.x. (See: https://github.com/tgsmith61591/pyramid/issues/16).

The fix is to add another build to the Travis matrix to build with cp27mu. I refactored the Linux portion of the `build_wheels.sh` to expose the UCS setting and avoid hardcoding it. The rest of the builds will default to UCS-2 (which was the default previously, too) and UCS-4 will only be built when explicitly set in a particular job's env. In short, this simply adds a new build without affecting existing builds.


tgsmith61591:
Thanks @shoelsch ! @charlesdrotar would you review when you get a chance?

Issue #39, opened 2018-10-30
OSError: Could not read saved model state from  Does it still exist?


sparo-jack:
/home/user/.pyramid-arima-cache/2018-10-30_23-34-59.732854-111--9223363280821728957.pmdpkl. Does it still exist?

tgsmith61591:
Please follow the instructions in the issue template! Can you provide the following:

* Pyramid version
* Command you're running

While you're at it, please take a look at this to understand serialization in Pyramid:

https://www.alkaline-ml.com/pyramid/serialization.html

If you're in an older version of pyramid (pre 0.8.1), you might be encountering an older issue where older cached states were scrapped if re-pickling.

tgsmith61591:
Haven't heard back on this issue, but it should have been handled in #42. If any issues persist, we can reopen...

Issue #38, opened 2018-10-15
Create Updated Wheels for Python 3.6+


DeastinY:
Seems to only be available for Python 3.6 on Mac :)

tgsmith61591:
Yeah, so this is something I mentioned at some point in #36, but the newest version will have wheels for 3.6 :)

DeastinY:
thanks a bunch! <3

tgsmith61591:
Handled in #42

Issue #37, opened 2018-10-11
How to auto get value d?


createmysoul:
hi
Could you tell me How to auto get value d?
thanks!

tgsmith61591:
It's not a silver bullet, but here are some tips and tricks in the user guide:

https://www.alkaline-ml.com/pyramid/tips_and_tricks.html#understanding-differencing

tgsmith61591:
It's also worth mentioning that `auto_arima` will estimate `d` if it's set to `d=None`. You can control the number of differencing terms it will try with the `max_d` parameter.

PR #42, opened 2018-11-02
Release 0.9.0

tgsmith61591:
v0.9.0
------

* Explicitly catch case in ``auto_arima`` where a value of ``m`` that is too large may over-estimate
  ``D``, causing the time series to be differenced down to an empty array. This is now handled by
  raising a separate error for this case that better explains what happened.

* Re-pickling an ``ARIMA`` will no longer remove the location on disk of the cached ``statsmodels``
  ARIMA models. Older versions encountered an issue where an older version of the model would be
  reinstated and immediately fail due to an OSError since the cached state no longer existed. This
  means that a user must be very intentional about clearing out the pyramid cache over time (addresses #39)

* Added pyramid cache check on initial import to warn user if the cache size has grown too large.

* If ``d`` or ``D`` are explicitly defined for ``auto_arima`` (rather than ``None``), do not
  raise an error if they exceed ``max_d`` or ``max_D``, respectively (sort of addresses #37)

* Added Circle CI for validating PyPy builds (rather than CPython)

* Deploy python wheel for version 3.6 on Linux and Windows (addresses #38)

* Include warning for upcoming package name change (``pmdarima``).


tgsmith61591:
@charlesdrotar I can always force merge, but curious for your thoughts (hence review request) on how the warning is raised in the `__init__.py` (removed in the next release with migration to `pmdarima`)

Issue #48, opened 2018-11-24
Could not read saved model state


jussiseppala:
#### Description
I have problem in loading the model from file when running the code on AWS Lambda.
The issue is that everything works fine when I'm doing this on my linux but it does not work on AWS Lambda. Meaning: 

- I'm able to create the model

- I'm able to store the model as file

- I'm able to load the model from file

- I'm also able to load the model from file which is located on AWS S3 bucket

But when I deploy my code to AWS Lambda using serverless framework loading the model fails. I expect that it is related to cache folder used in pmdarima.

#### Steps/Code to Reproduce
This the code which loads the model from AWS S3 bucket.

```python
import boto3
import botocore
import os
import io
import numpy as np
from io import BytesIO
from sklearn.externals import joblib

BUCKET_NAME = 'mybcucket'

# downloads model from S3 into give file(name)
def download_model_aws(to_file_name):
	bucket = boto3.resource("s3").Bucket(BUCKET_NAME)
	with BytesIO() as modelfo:
		bucket.download_fileobj(Key=to_file_name, Fileobj=modelfo)
		model = joblib.load(modelfo)
	return model

file_name = 'forecast_models/%d-forecast.pkl' % company_id
model = download_model_aws(file_name)
```

#### Expected Results

The model should be loaded from file successfully.

#### Actual Results

Following exception is thrown:

```
Could not read saved model state from /home/myuser/.pyramid-arima-cache/2018-11-24_07-42-29.725501-000--9223363283404616476.pmdpkl. Does it still exist?: OSError
Traceback (most recent call last):
...
File "/var/task/main.py", line 46, in get_model
return download_model_aws(file_name)
File "/var/task/main.py", line 31, in download_model_aws
model = joblib.load(modelfo)
File "/var/task/sklearn/externals/joblib/numpy_pickle.py", line 568, in load
obj = _unpickle(fobj)
File "/var/task/sklearn/externals/joblib/numpy_pickle.py", line 508, in _unpickle
obj = unpickler.load()
File "/var/lang/lib/python3.6/pickle.py", line 1050, in load
dispatch[key[0]](self)
File "/var/task/sklearn/externals/joblib/numpy_pickle.py", line 328, in load_build
Unpickler.load_build(self)
File "/var/lang/lib/python3.6/pickle.py", line 1507, in load_build
setstate(state)
File "/var/task/pmdarima/arima/arima.py", line 587, in __setstate__
'Does it still exist?' % loc)
[...]

PR #58, opened 2018-12-03
[MRG] Re-design ARIMA serialization

tgsmith61591:
This PR addresses #48.

### The problem

There was [a bug](https://github.com/statsmodels/statsmodels/issues/3401) in older versions of statsmodels that prevented SARIMAX models from being pickled in Python 3.6. As a result, we originally had some fancy, custom pickling magic that separately saved the statsmodels results wrappers in the `pmdarima` cache. When deserializing models, the cached results wrapper was loaded from disk. The problem is this reduced model portability, and when moving a serialized model to a separate environment, attempting to load it from disk would result in an `OSError` when the cached state couldn't be found.

### The solution

Since we now require statsmodels 0.9.0, we no longer need the automagical pickling hack, and we can simply resort to bundling the entire `ARIMA` object into one pickle file. This also means the `PMDARIMA_CACHE` and `PMDARIMA_CACHE_WARN_SIZE` environment variables are no longer needed, but will be maintained for legacy's sake.


charlesdrotar:
LGTM!

PR #60, opened 2018-12-03
[MRG + 1] Generate plot diff rmse

charlesdrotar:
Add the ability to calculate diffs of plots so changes in figures can be tested as well as the execution. This addresses [Issue#59](https://github.com/tgsmith61591/pmdarima/issues/59).


pmdarima/arima/tests/test_arima.py
```diff
@@ -24,6 +24,25 @@
...
+
+    # Only do this setup if we are not running on Travis CI.
+    from matplotlib.pyplot import savefig
+    from matplotlib.testing.exceptions import ImageComparisonFailure
+    import matplotlib._png as _png
+
+    test_images_input_path = \
+        os.path.join(os.path.dirname(__file__), 'input_images')
+    test_images_output_path = \
+        test_images_input_path.replace('input_images', 'output_images')
```
tgsmith61591: Can you wrap these in `try`/`finally` so that we always nuke the .png files at the end?

pmdarima/arima/tests/test_arima.py
```diff
@@ -24,6 +24,25 @@
...
+
+if not travis:
+
+    # Only do this setup if we are not running on Travis CI.
+    from matplotlib.pyplot import savefig
+    from matplotlib.testing.exceptions import ImageComparisonFailure
+    import matplotlib._png as _png
+
+    test_images_input_path = \
+        os.path.join(os.path.dirname(__file__), 'input_images')
```
tgsmith61591: This is why you're failing on Appveyor. The package is installed to site-packages, and this is going to look for the input_images path in site-packages, which fails. It's OK (and actually probably better) to use relative paths for something like this

tgsmith61591:
@charlesdrotar, can you add `matplotlib` to the `test-requirements` section of the `Makefile`? @aaronreidsmith pointed that out earlier .. we no longer require it to build, but we do need it for test. Though the CI envs explicitly install it, we should automate that in the make stage

pmdarima/arima/tests/test_arima.py
charlesdrotar: So I'm not a big fan of the relative path solution to look for the input images because you have to run from a certain path in order for the tests to work and pull in the images correctly and that path has to be the same for each CI environment. That would mean running AppVeyor and Circle CI would either both have to be executed from `empty_folder` since AppVeyor is using that or some commonly defined location. If I wrote the tests to look at that from a path in AppVeyor it would not work in CircleCI because the relative path would be different for that reason. It would mean mirroring work across all CI solutions in order for execution to be the same.

So as far as `site-packages` is concerned, I verified that is the case with both the Circle and AppVeyor installations that it is put in `site-packages`,  but the tests did execute and pass for Circle and not for AppVeyor, which to me was a difference in the setup/installation process as to why one would have visibility and not the other. So I decided to dig further and I think I have an exact answer :smiley:.

To fully understand the nature of this error, I tested pip-installing on the built artifacts, `pip install --pre --no-index --find-links dist/ pmdarima`,   and pip-installing using `pip install --editable pmdarima`. I believe the root cause for why it fails singly within AppVeyor as opposed to CircleCI as well is we are using the artifacts within the `dist/` directory to do the install of pmdarima for AppVeyor, and within Circle we are doing `pip install --editable pmdarima` which allows us to see the input images because the artifacts and the input images are colocated.

See excerpt from **appveyor.yml**:

```YAML

  # 9/25/18 - Use the scikit approach

  # Build the package (and the wheels!)

  - "%CMD_IN_ENV% pip install --timeout=60 -r build_tools/appveyor/requirements.txt"

  - "%CMD_IN_ENV% python setup.py bdist_wheel bdist_wininst" # <---- <<THIS LINE>>
  - ps: "ls dist"



  # Install the generated wheel file

  - "pip install --pre --no-index --find-links dist/ pmdarima" # <---- <<THIS LINE>>




# Not a .NET project, we build pmdarima in the install step instead

build: false



test_script:

  # Change to a non-source folder to make sure we run the tests on the

  # installed library

  - mkdir "../empty_folder"

  - "cd ../empty_folder"



  # Test with pytest

  - "pytest --showlocals --durations=20 --pyargs pmdarima"

  - "cd ../pmdarima"
```

That said, we should just be able to switch `pip install --pre --no-index --find-links dist/ pmdarima` with `pip install --editable pmdarima` and it should work. I verified this behavior locally to replicate the error and the workaround using the proposed solution and the artifacts are still built so this shouldn't affect the CD portion. Not sure if there is any reason why we can't point at the installed code directly for the tests so we have visibility to the image files. 

@tgsmith61591 Thoughts?

tgsmith61591:
@charlesdrotar any update on this?

charlesdrotar:
Unfortunately not at this point.

tgsmith61591:
Interesting that you ran into the `LinAlgError` here... did you change any version requirements? We're running into this in #74 for Py3.7 support and still can't figure out why

pmdarima/arima/tests/test_arima.py
charlesdrotar: Done!

charlesdrotar:
@tgsmith61591 @aaronreidsmith The code the there and the tests pass, but the coverage is dropping because we can't test the matplotlib plotting in Travis and that is unfortunately where coverage is getting tested from. Not sure if this needs to be addressed at this moment or not.

tgsmith61591:
Coverage is reported from circle as of recently. You up to date on `develop`?

pmdarima/arima/tests/test_arima.py
```diff
@@ -16,14 +16,34 @@
...
+    import matplotlib._png as _png
+
+    test_images_input_path = \
+        os.path.join('pmdarima', 'arima', 'tests', 'input_images')
+    test_images_output_path = \
+        test_images_input_path.replace('input_images', 'output_images')
+
+    # make output directory if it doesn't exist
+    if not os.path.exists(test_images_output_path):
+        os.makedirs(test_images_output_path)
```
tgsmith61591: Can you put this whole `if not travis` block inside the other? And this portion in particular should go inside the `try/finally`

pmdarima/arima/tests/test_arima.py
```diff
@@ -912,3 +916,154 @@ def test_new_serialization():
...
+        """
+
+        if expected_image.shape != actual_image.shape:
+            raise ImageComparisonFailure(
+                "Image sizes do not match expected size: {} "
+                "actual size {}".format(expected_image.shape,
+                                        actual_image.shape))
+        # Convert to float to avoid overflowing finite integer types.
+        return np.sqrt(((expected_image - actual_image).astype(float) ** 2)
+                       .mean())
```
tgsmith61591: This is great work man!

tgsmith61591:
@charlesdrotar Everything LGTM. I do have one tiny little nitpick I commented on, but it's tiny. Also, looks like the PNG files are still in the files to merge--can you remove those?

Issue #54, opened 2018-11-29
Feature Request: Model output as a Dictionary


Jacyle:
#### Description
By outputting the model summary statistics, it will allow for passing descriptive statistics like p-value, R-squared, coefficients, the model specifications (ex. 'order = (1,1,1)'), etc... as parameters to other parts of your code. This can be useful for setting logic and thresholds for model requirements (ex. if r_sq >= 0.80).

tgsmith61591:
Problem setup:

```python

from pmdarima.datasets import load_lynx
from pmdarima.arima import auto_arima
from sklearn.metrics import mean_squared_error
import matplotlib.pyplot as plt
import numpy as np

# Load the data and split it into separate pieces
data = load_lynx()
train, test = data[:90], data[90:]

# Fit a simple auto_arima model
modl = auto_arima(train, start_p=1, start_q=1, start_P=1, start_Q=1,
                  max_p=5, max_q=5, max_P=5, max_Q=5, seasonal=True,
                  stepwise=True, suppress_warnings=True, D=10, max_D=10,
                  error_action='ignore')
```

Are you looking for something like this?

```python
>>> {'pvalues': modl.pvalues(), 'resid': modl.resid(), 'order': modl.order,
     'seasonal_order': modl.seasonal_order, 'oob': modl.oob(), 'aic': modl.aic(),
     'aicc': modl.aicc(), 'bic': modl.bic(), 'bse': modl.bse(), 'params': modl.params()}
{'pvalues': array([2.04752445e-03, 1.43710465e-61, 1.29504002e-10, 5.22119887e-15]), 'resid': array([-1244.3973072 ,  -302.89697033,  -317.63342593,  -304.57267897,
         131.69413491,   956.15566697,   880.37459722,  2445.86460353,
        -192.84268876,  -177.1932523 ,  -101.67727903,   384.05487582,
        -304.52047818,  -570.72748088,  -497.48574217,  1286.86848903,
        -400.22840217,  1017.55518758, -1157.37024626,  -295.26213543,
         104.79931827,  -574.9867485 ,  -588.49652697,  -535.37707505,
        -355.71298419,  -164.06179682,   574.51900799,    15.45522718,
       -1358.43416826,   120.42735893,  -147.94038284,  -685.64124874,
        -365.18947057,  -243.79704985,   317.79437422,   585.59553667,
          34.70605783,  -216.21587989,  -692.53375089,   116.87379358,
[...]

PR #61, opened 2018-12-04
[MRG+1] Added to_dict to ARIMA class

aaronreidsmith:
#54 Added `to_dict` method to ARIMA class.

The tests gave me trouble. I wanted to test each key-value pair individually, but they all had to be treated slightly differently (hence the multiple asserts). If anyone has suggestions to clean that test up, please let me know.

### DIFF
--- a/pmdarima/arima/tests/test_arima.py
+++ b/pmdarima/arima/tests/test_arima.py
@@ -797,7 +797,7 @@ def test_with_intercept(order, seasonal):
 
 
 def test_to_dict_returns_dict():
-    train, test = lynx[:90], lynx[90:]
+    train = lynx[:90]
     modl = auto_arima(train, start_p=1, start_q=1, start_P=1, start_Q=1,
                       max_p=5, max_q=5, max_P=5, max_Q=5, seasonal=True,
                       stepwise=True, suppress_warnings=True, D=10, max_D=10,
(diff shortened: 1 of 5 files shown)


aaronreidsmith:
#54 Forgot to reference the issue...

pmdarima/arima/tests/test_arima.py
```diff
@@ -801,3 +801,78 @@ def test_with_intercept(order, seasonal):
...
+        'oob': np.nan,
+        'aic': 1487.8850037609368,
+        'aicc': 1488.5992894752226,
+        'bic': 1497.8842424422578,
+        'bse': np.array([2.26237893e+02, 6.97744631e-02,
+                         9.58556537e-02, 1.03225425e+05]),
+        'params': np.array([6.97548186e+02, 1.15522102e+00,
+                            -6.16136459e-01, 8.07374077e+05])
+    }
+
```
tgsmith61591: One way you could handle the assertion of the values is a loop with `getattr`. Since you're not really asserting the values themselves (their methods have been extensively tested) but that the attribute is mapped to the appropriate name, you could do something like:

```python
for k, v in actual.items():
    attr = getattr(modl, k)  # might be a method
    if hasattr(attr, "__call__"):
        attr = attr()
    assert_array_almost_equal(v, attr)
```

Just a thought... @charlesdrotar - thoughts?

charlesdrotar: I would use [assertDictEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertDictEqual). It is part of base python and will give more informative error messages on diffs.

tgsmith61591: I like that. We can add that into `pmdarima/utils/testing` in a separate feature and replace this assertion logic with that function

tgsmith61591:
@aaronreidsmith had to merge `develop` into your branch to get it to build on CI since there was another PR merged overnight. There are linting errors in the tests, FYI. Make sure you pull your branch before fixing:

```bash
$ git pull origin arima_to_dict_54
```

aaronreidsmith:
Sorry for the bad commit history on this one. I think this should be good now. I agree with @tgsmith61591 that cleaning up the fuzzy dictionary comparison may be a future issue.

Issue #62, opened 2018-12-06
Versioned documentation


tgsmith61591:
## Current state

As of version 1.0.0, we're only building/deploying documentation on master, and only for the latest version. See:

http://alkaline-ml.com/pmdarima

## Ideal state

I'd like us to start building versioned documentation such that every new release becomes `latest`, and all previous releases land in a versioned subdirectory of the documentation website. This happened by default when we hosted on `readthedocs`, but with our more new, more complicated build, and desire for a custom domain, RTD wasn't cutting it.

The solution will entail:

* [x] Latest release doc deployed to `alkaline-ml.com/pmdarima/`
* [x] Latest release doc deployed to `alkaline-ml.com/pmdarima/<version>/`
* [x] We build documentation also for `develop` located at `alkaline-ml.com/pmdarima/develop/`

## Steps to solve

1. When we build `gh-pages` releases on Circle, we'll move the existing HTML artifacts into a new subdirectory called "`<VERSION>`", which means we'll need to land a `VERSION` file in the branch for every build on `master`.
2. Every `develop` build will re-deploy the `develop` ("bleeding edge") doc

As an added feature, I'd like us to include the last two-to-three version of both `pmdarima` and `pyramid-arima` for those legacy users out there. This would be a one-time build & push from the existing release branches, and would later be automated.

tgsmith61591:
@charlesdrotar & @aaronreidsmith tagging you guys to generate some discussion. But I will probably subtask this. This would be a (*very*) nice-to-have prior to the 1.1.0 release.

PR #69, opened 2018-12-16
[MRG] Versioned documentation

tgsmith61591:
This solves #62 

## Automated:

* Latest release doc will be deployed to `alkaline-ml.com/pmdarima/`
* Latest release doc will be deployed to `alkaline-ml.com/pmdarima/<version>/`
* We build documentation also for `develop` located at `alkaline-ml.com/pmdarima/develop/`

I also manually added versioned doc for [v0.9.0](http://www.alkaline-ml.com/pmdarima/0.9.0/) and [v1.0.0](http://www.alkaline-ml.com/pmdarima/1.0.0/). Eventually we can link to older versions from the home page.

I'm particularly interested in a review of `build_tools/circle/build_push_doc.sh`

## Unrelated:

I also merged another branch (`improve-cython-speed`) into this branch at one point for some ancillary, related reason, so that branch is also getting merged...

### DIFF
--- a/build_tools/circle/build_push_doc.sh
+++ b/build_tools/circle/build_push_doc.sh
@@ -56,19 +56,21 @@ if [[ ${CIRCLE_BRANCH} == "master" || ${CIRCLE_BRANCH} == "develop" || ${CIRCLE_
 
   # On both of these, we'll need to remove the artifacts from the package
   # build itself
-  declare -a leftover=(".cache"
-                       ".idea"
+  declare -a leftover=(".pyramid-cache"
+                       ".pytest_cache"
+                       "benchmarks"
                        "build"
-                       "build_tools"
+                       "dist"
                        "doc"
-                       "examples"
                        "pmdarima"
-                       "pmdarima.egg-info")
+                       "pmdarima.egg-info"
+                       "pyramid_arima.egg-info")
 
   # check for each left over file/dir and remove it
   for left in "${leftover[@]}"
   do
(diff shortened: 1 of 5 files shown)


build_tools/circle/build_push_doc.sh
```diff
@@ -32,50 +32,114 @@ git pull origin gh-pages
...
 
-# Add everything, get ready for commit. But only do it if we're on master
-if [[ "${CIRCLE_BRANCH}" =~ ^master$|^[0-9]+\.[0-9]+\.X$ ]]; then
-    git add --all
-    git commit -m "[ci skip] publishing updated documentation..."
+# Otherwise it's master or another branch, which is a bit more involved.
+# Remove the artifacts from the previous deployment, move the new ones into the
+# folder as well as into the versioned folder. This won't be deployed unless
+# it's master
+else
```
aaronreidsmith: Maybe this is naïve, but do we want to go through this whole process if it is not `master` or `develop`? I know we don't deploy, but we could theoretically catch this here and not build docs for feature branches at all.

tgsmith61591: Yes, and actually originally I did exactly that (ignored for other branches), but it occurred to me we want to make sure new PRs will still pass on all this logic before we let them into `develop` or `master` (particularly if we forget to bump a version or anything)

aaronreidsmith: Okay, that makes sense!

Issue #67, opened 2018-12-14
Clarify ADFTest


Juanlu001:
With statsmodels I get the same values as with R:

```r
> adf.test(c(1, 2, 1, 2.1, 2, 2, 1, 2, 1))
Augmented Dickey-Fuller Test 
alternative: stationary 
 
Type 1: no drift no trend 
     lag    ADF p.value
[1,]   0 -0.706   0.417
[2,]   1 -0.303   0.548
[3,]   2 -0.146   0.593
Type 2: with drift no trend 
     lag   ADF p.value
[1,]   0 -4.81   0.010
[2,]   1 -1.39   0.547
[3,]   2 -1.67   0.450
Type 3: with drift and trend 
     lag   ADF p.value
[1,]   0 -4.18  0.0169
[2,]   1 -1.18  0.8843
[3,]   2 -1.02  0.9183
---- 
Note: in fact, p.value = 0.01 means p.value <= 0.01
```

```python
>>> adfuller([1, 2, 1, 2.1, 2, 2, 1, 2, 1], maxlag=0, regression='c', autolag=None)[:2]
(-4.807115469086105, 5.270676650320141e-05)
>>> adfuller([1, 2, 1, 2.1, 2, 2, 1, 2, 1], maxlag=1, regression='c', autolag=None)[:2]
(-1.3941478052336627, 0.5850784962155383)
>>> adfuller([1, 2, 1, 2.1, 2, 2, 1, 2, 1], maxlag=2, regression='c', autolag=None)[:2]
(-1.6697993873971613, 0.44673603965371517)
```

However, with `pmdarima.ADFTest` I'm not sure what am I getting:

```python
>>> ADFTest(alpha=0.05, k=2).is_stationary(np.array([1, 2, 1, 2.1, 2, 2, 1, 2, 1]))
(0.3718647834033668, False)
>>> ADFTest(alpha=0.05, k=2).is_stationary(np.array([1, 2, 1, 2.1, 2, 2, 1, 2, 1]))
(0.5754585996166532, False)
```

Juanlu001:
Awesome, thank you very much! :)

PR #71, opened 2018-12-18
[MRG] Greater stationarity test precision

tgsmith61591:
Issue #67 asked for clarification around the use of the stationarity tests in `pmdarima`. In the process of putting together an example, I found a slight bug in the way we compute `OLS` regression for the augmented Dickey-Fuller test, as well as several other opportunities for improvement. This PR achieves the following:

* Change the `OLS` fit method to use `'qr'` rather than `'pinv'`
* Use the `OLS` result wrapper `bse` attribute to access coefficient standard errors rather than the `H0_se` attribute
* Improve readability in the `KPSS` test
* Include links to R source code in the stationarity test docstrings
* Add several tests and more coverage

### DIFF
--- a/pmdarima/arima/stationarity.py
+++ b/pmdarima/arima/stationarity.py
@@ -43,6 +43,9 @@ class _BaseStationarityTest(six.with_metaclass(ABCMeta, BaseEstimator)):
     def _embed(x, k):
         # lag the vector and put the lags into columns
         n = x.shape[0]
+        if k > n:
+            raise ValueError("k cannot exceed y dim")
+
         rows = [
             # so, if k=2, it'll be (x[1:n], x[:n-1])
             x[j:n - i] for i, j in enumerate(range(k - 1, -1, -1))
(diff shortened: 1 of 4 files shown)


pmdarima/arima/stationarity.py
```diff
@@ -172,10 +175,14 @@ def is_stationary(self, x):
         eta = (s * s).sum() / (n**2)
         s2 = (e * e).sum() / n
 
-        scalar, denom = 10, 14
+        # scalar, denom = 10, 14
+        # if self.lshort:
+        #     scalar, denom = 3, 13
```
aaronreidsmith: Is this portion being left commented out intentionally?

tgsmith61591: Yep. Replaced by the new code:

```python
if self.lshort:
    l_ = int(np.trunc(4 * (n / 100) ** 0.25))
else:
    l_ = int(np.trunc(12 * (n / 100) ** 0.25))
```

tgsmith61591: Left the old for posterity's sake 😉

aaronreidsmith: Gotcha. I figured, but thought I would ask!

Issue #76, opened 2018-12-21
Python 3.7 Travis builds produce no artifacts


tgsmith61591:
#### Description

#74 is nearly resolved and we're almost ready to merge #75 into `develop`. But the Travis build currently doesn't produce a wheel artifact for the Python 3.7 build on Linux. Look at [this job](https://travis-ci.org/tgsmith61591/pmdarima/jobs/470774964), for example, and click on the arrow next to `ls dist/`. On other builds, we see contents. Take the Mac 3.7 build, for instance:

```bash
$ ls dist/
pmdarima-1.1.0.dev0-cp37-cp37m-macosx_10_7_x86_64.whl
pmdarima-1.1.0.dev0.tar.gz
```

#### Expected Results

We should expect the Travis build for 3.7 on Linux to produce both a `sdist` tar and a wheel file, verifiable by viewing the contents of `dist/`.

#### Threads to pull at

I suspect the culprit is somewhere in `.travis.yml`, or `build_tools/travis/build_wheels.sh` (maybe even `build_tools/travis/build_manywheels_linux.sh` since it specifies a lower version of Cython).

Can someone branch the `py37` branch and look into this? I'm OK getting `py37` into develop without this, knowing that this may take us down a bit of a rabbit trail, but this needs to be done prior to the 1.1.0 release, as otherwise we won't end up distributing wheels for 3.7!

@aaronreidsmith may be a good place to jump in and learn the Travis ropes

aaronreidsmith:
I can look into this. Now that `py37` is in `develop`, can I just branch off `develop`?

aaronreidsmith:
Okay, 0.27.2 made it fail with a different error [see here](https://travis-ci.org/tgsmith61591/pmdarima/jobs/471199629). I will try `>=0.28`

tgsmith61591:
You may even just drop the version and let it install the latest.

aaronreidsmith:
So, it looks like the latest build for 3.7 did produce wheels ([link](https://travis-ci.org/tgsmith61591/pmdarima/jobs/471202542#L2471)). As far as the `DEPLOY=true` assertion, where would that go? Kind of new to this...

tgsmith61591:
I'd create a new bash script in `build_tools/travis` called `after_wheel_build.sh` (or whatever) and assert the 'whl' pattern is present in the `dist/` directory. Make sure you run that job after (or in place of) the `ls dist/` command. 

[...]

PR #77, opened 2018-12-23
[MRG+1] Py37 build artifacts

aaronreidsmith:
This PR removes the explicit version for `cython`, which was preventing build artifacts for Python 3.7 (#76), and adds a check to make sure our build artifacts were created before deployment.

I added the check in the `before_deploy` section of the build as noted [here](https://docs.travis-ci.com/user/job-lifecycle/#deploying-your-code), so it will only get run on a tagged release (i.e., it won't run when we merge this in). I initially had it in the `after_success` phase, and I am more than happy to move it back, but it made more sense to me here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,7 +96,7 @@ after_success:
   - source build_tools/travis/after_success.sh  # || echo "shell_session_update failed"
   # Build the wheels every time so we can debug
   - bash build_tools/travis/build_wheels.sh
-  - ls dist/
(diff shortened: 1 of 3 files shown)


aaronreidsmith:
If y'all would like to see a build where the assertion ran correctly (since it won't run unless we are on a tagged branch), see [here](https://travis-ci.org/tgsmith61591/pmdarima/jobs/471637401#L2471)

Issue #88, opened 2018-12-31
Add more tests of seasonality


tgsmith61591:
#### Description

Right now, the only test of seasonality we implement is the Canova-Hansen test (`CHTest`), which is where the complaints of slowness come from (see #84).

#### Steps/Code to Reproduce

The CH test is known not to be highly performant. Even R will choke on large datasets when the test is set to "ch", where `pmdarima` will power through in < 10 seconds. Using the (very large) dataset given in issue #32, here is some proof:

#### R's `auto.arima` (could not get to finish):

`nsdiffs` comes back, but super slowly (several minutes):

```R
library(forecast)
X = read.csv('item_sales_daily.csv.gz')
y = ts(X$sales, frequency=365)
nsdiffs(sales, test='ch', max.D=3)
```

As for `auto.arima` itself, it takes FOREVER.

```R
# Go pick up a brisket, smoke it low and slow overnight and then
# come back to see if this bad boy has finished:
auto.arima(y, seasonal=TRUE, seasonal.test='ch')
```

#### `pmdarima` (still very slow):

`nsdiffs` executes in < 10 seconds:

```python
import pandas as pd
import pmdarima as pm

X = pd.read_csv('item_sales_daily.csv.gz')
y = X['sales'].values

pm.arima.nsdiffs(y, m=365, test="ch", max_D=3)
```

The same exact `auto_arima` code in `pmdarima` executes very slowly, but it finishes. On my machine, I got a result in several minutes.

```python
pm.auto_arima(y, m=365, stepwise=True, seasonal_test='ch')
```

#### Versions affected

All versions of `pmdarima` < 1.2.0

### Solution(s)

We need to implement more tests of seasonality. Good starts would be:

* [x] OCSB test
* [ ] HEGY test

tgsmith61591:
After #121, I can get this to fit in about 3 minutes now.

```python
import pandas as pd
import pmdarima as pm
from pmdarima.pipeline import Pipeline
from pmdarima.preprocessing.exog import FourierFeaturizer

X = pd.read_csv('item_sales_daily.csv.gz')
y = X['sales'].values
m = 365

pipe = Pipeline([
    ('fourier', FourierFeaturizer(m=m, k=20)),
    ('arima', pm.AutoARIMA(m=m, stepwise=True, seasonal=False))
])

pipe.fit(y)
```

Comparatively, the equivalent R code runs in about 1.5 mins:

```R
library(forecast)
[...]

PR #89, opened 2019-01-01
[MRG] Improve speed of seasonal tests

tgsmith61591:
* Increase speed of `CHTest` Cython computation
* Add `OCSBTest`, which is orders of magnitude faster and addresses part of #88 
* Make "ocsb" default for `nsdiffs` and `auto_arima`
* Add seasonal test benchmarking notebook


tgsmith61591:
@charlesdrotar after this you can add your benchmarking feature :)

tgsmith61591:
@charlesdrotar TBH I don't really love the idea of having 25-minute tests running... Rather than benchmarking the known slowest tests, can we benchmark a set that are critical in speed? I.e., anything over a certain time limit will fail

pmdarima/tests/test_benchmarks.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def test_oob_for_issue_29(self, benchmark):
+        benchmark(test_arima.test_oob_for_issue_29)
+
+    def test_oob_sarimax(self, benchmark):
+        benchmark(test_arima.test_oob_sarimax)
+
+    @pytest.mark.parametrize('m', ch_test_parameters)
+    def test_ch_test_long(self, benchmark, m):
+        benchmark(test_seasonality.test_ch_test_long, m=m)
```
tgsmith61591: If I simply run `pytest`, is it going to run this?

charlesdrotar: Yes, unless you tell it to skip benchmark tests.

PR #94, opened 2019-01-21
[MRG+1] Deprecate 'is_stationary'

tgsmith61591:
The tests of stationarity test currently use a function "`is_stationary`" to indicate whether we should difference the time series. However, the nomenclature is backwards. Since we've recently had an issue where someone requested clarification on the test (#67), it makes sense we should change this so it makes more intuitive sense.

### Changes

* Create new function, "`should_diff`", that replaces `is_stationary`
* If user calls `is_stationary`, raise a `DeprecationWarning` and internally delegate to `should_diff`

We'll remove the `is_stationary` function in 1.4.

### DIFF
--- a/doc/tips_and_tricks.rst
+++ b/doc/tips_and_tricks.rst
@@ -150,7 +150,7 @@ for more info. Here's an example of an ADF test:
     # Test whether we should difference at the alpha=0.05
     # significance level
     adf_test = ADFTest(alpha=0.05)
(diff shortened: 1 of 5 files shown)


tgsmith61591:
@aaronreidsmith No, the new Numpy release creates a litany of stupid artifacts that are breaking the `gh-pages` deployment. I have an idea of what's going on... and it's what's causing all PRs into `develop` to break.

:shaking-fist: Curse you, Numpy developers!

PR #97, opened 2019-02-07
Less appveyor

tgsmith61591:
Our Appveyor builds are prohibitively slow. We still need all jobs to run since for deployments the wheel has to be built on the appropriate architecture... but we don't always have to run tests.

This PR removes half of the test jobs, just letting the wheels build. It also adds in some more pip caching, although I'm wary as to whether it's actually working properly or not.

#### Bonus

Fix more pypy breakage on Circle. Any subsequent PR will have to rebase off these changes once they make it into `develop`

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -61,6 +61,10 @@ environment:
       PYTHON: "C:\\Python37-x64"
       DO_TEST: "true"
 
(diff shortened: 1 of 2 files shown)


tgsmith61591:
Merged as admin

Issue #112, opened 2019-03-27
README formatting messed up on PyPI


aaronreidsmith:
The formatting for the README is getting screwed up when we release to PyPI ([example](https://pypi.org/project/pmdarima/)). This has only been an issue since `1.1.0`, [`1.0.0`'s formatting](https://pypi.org/project/pmdarima/1.0.0/) looks fine.

I suspect this has something to do with our Markdown being interpreted as RST. I will look into this

aaronreidsmith:
@tgsmith61591 do we explicitly install `readme_renderer[md]` before distribution? Based on my local testing, I feel like that is the issue.

I know you made the `make` target, but it only relies on `bdist_wheel` (and by extension `requirements`), so `twine` doesn't appear to be checked for before testing the README.
 
Before `readme_renderer[md]`:
```
(pmdenv) asmith ~/Documents/projects/pmdarima (pypi-docs) $ twine check dist/*
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/readme_renderer/markdown.py:38: UserWarning: Markdown renderers are not available. Install 'readme_render[md]' to enable Markdown rendering.
  warnings.warn(_EXTRA_WARNING)
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/readme_renderer/markdown.py:50: UserWarning: Markdown renderers are not available. Install 'readme_render[md]' to enable Markdown rendering.
  warnings.warn(_EXTRA_WARNING)
Checking distribution dist/pmdarima-1.2.0.dev0-cp36-cp36m-macosx_10_7_x86_64.whl: Failed
The project's long_description has invalid markup which will not be rendered on PyPI. The following syntax errors were detected:
Checking distribution dist/pmdarima: InvalidDistribution: Unknown distribution format: 'pmdarima'
(pmdenv) asmith ~/Documents/projects/pmdarima (pypi-docs) $ twine check dist/*
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/readme_renderer/markdown.py:38: UserWarning: Markdown renderers are not available. Install 'readme_render[md]' to enable Markdown rendering.
  warnings.warn(_EXTRA_WARNING)
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/readme_renderer/markdown.py:50: UserWarning: Markdown renderers are not available. Install 'readme_render[md]' to enable Markdown rendering.
[...]

PR #116, opened 2019-03-28
[MRG+1] PyPI README 

aaronreidsmith:
This PR adresses #112.

I believe the only issue was the lack of `readme_renderer[md]`. I have added it to the `twine-check` make target, as well as the Travis and Appveyor requirements.


tgsmith61591:
After making one more quick addition to the `Makefile`, looks like this is working on PyPi test:

https://test.pypi.org/project/pmdarima/

That is, when deploying from a Mac... looks like the last two deployments have made it to PyPi from Appveyor first. Wonder if that has anything to do with it?

Makefile
```diff
@@ -12,6 +12,16 @@ clean:
...
+deploy-requirements:
+	$(PYTHON) -m pip install twine readme_renderer[md]
+
+# Depends on an artifact existing in dist/, and two environment variables
+deploy-twine-test: bdist_wheel deploy-requirements
+	$(PYTHON) -m twine upload \
+		--repository-url https://test.pypi.org/legacy/ dist/* \
+		--username ${TWINE_USERNAME} \
+		--password ${TWINE_PASSWORD}
+
```
tgsmith61591: This requires you to have these in your environment, obviously, or it won't work.

aaronreidsmith:
> That is, when deploying from a Mac... looks like the last two deployments have made it to PyPi from Appveyor first. Wonder if that has anything to do with it?

That may actually be a factor, according to https://github.com/pypa/setuptools/issues/1440, the Windows newlines can screw things up. Wonder if there is a way to account for that on Appveyor

tgsmith61591:
I think that's the core problem. Maybe we can account for that in this line of the `setup.py`:

```python
with open("README.md", "r") as fh:
    LONG_DESCRIPTION = fh.read()
```

Maybe something like

```
# Alert, this is un-tested...
with open("README.md", "r") as fh:
    LONG_DESCRIPTION = fh.read().replace(os.linesep, "\n")
```

aaronreidsmith:
[This is the issue](https://ci.appveyor.com/project/tgsmith61591/pmdarima/builds/23441425/job/nq2lw7fw2qwnlw8k#L521)

```
Checking distribution dist\pmdarima-1.2.0.dev0-cp35-cp35m-win32.whl: warning: `long_description_content_type` missing.  defaulting to `text/x-rst`.
```

I will look into that before we merge this

aaronreidsmith:
So it looks like Appveyor had an old version of `setuptools`:

```
Requirement already satisfied: setuptools in c:\python35\lib\site-packages (from pytest->-r build_tools/appveyor/requirements.txt (line 9)) (28.8.0)
```

Travis has the latest:

```
setuptools-40.2.0    | 571 KB    |            |   0% 
setuptools-40.2.0    | 571 KB    | ########8  |  88% 
setuptools-40.2.0    | 571 KB    | ########## | 100% 
```

because of this line in `build_tools/travis/build_manywheels_linux.sh`

```
${PIP} install --upgrade setuptools
```

So, after upgrading, here is what I get on Appveyor:

```
%CMD_IN_ENV% python -m twine check dist/*
Using default MSVC build environment for 32 bit architecture
Executing: python -m twine check dist/*
Checking distribution dist\pmdarima-1.2.0.dev0-cp35-cp35m-win32.whl: Passed
```

(based on [this SO post](https://stackoverflow.com/a/26737258))

appveyor.yml
```diff
@@ -101,6 +101,9 @@ install:
   - "%CMD_IN_ENV% python setup.py bdist_wheel"
   - ps: "ls dist"
 
+  # Check for README compatibility in case we deploy
+  - "%CMD_IN_ENV% python -m twine check dist/*"
```
tgsmith61591: If this doesn't pass, will the CI job fail?

tgsmith61591: (it should, probably)

build_tools/appveyor/requirements.txt
```diff
@@ -9,6 +9,8 @@ patsy
 pytest
 pytest-mpl
 pytest-benchmark
+setuptools>=38.6.0
```
tgsmith61591: 🙌

setup.py
```diff
@@ -31,7 +31,7 @@
 
 # Get the long desc
 with open("README.md", "r") as fh:
-    LONG_DESCRIPTION = fh.read()
+    LONG_DESCRIPTION = fh.read().replace(os.linesep, '\n')
```
tgsmith61591: Do we still need this? Maybe good to keep it in case setuptools goes crazy again :)

aaronreidsmith: I actually think this line is useless. When the file is read, it will either always have Unix newlines or Python can handle it (if someone changed the line endings and pushed for some reason...). The error in the issue I linked above happened at WRITE time (while building), which we can’t really control. So I will probably remove this before I merge.

appveyor.yml
aaronreidsmith: It does not at the moment. When we hit this line with the old `setuptools` version, it gave a warning and continued. I’ll figure out a way to make the job fail if this step fails

setup.py
tgsmith61591: @aaronreidsmith Roger that 💯

aaronreidsmith:
Okay, so we can see that the [Appveyor build will fail if the README renders wrong](https://ci.appveyor.com/project/tgsmith61591/pmdarima/builds/23455796/job/xt7q5a2p0o25r90t#L534) now. After re-upgrading `setuptools`, [it passes](https://ci.appveyor.com/project/tgsmith61591/pmdarima/builds/23456134/job/4y4alq3ad7o1kt7a#L531).

build_tools/travis/after_wheel_build.sh
```diff
@@ -9,3 +9,11 @@ else
     echo "Build artifacts were not created. Skipping deployment..."
     exit 1
 fi
+
+# If we get to this point, the build worked, and we need to check for README rendering
```
aaronreidsmith: @tgsmith61591 I added this check to Travis too, but it will only run when we are deploying. I don't have a good way to test it, so if we want to pull it out for this PR, that's fine by me

appveyor.yml
aaronreidsmith: This will fail on improper rendering now

PR #121, opened 2019-04-06
Add Fourier featurizer and Pipeline

tgsmith61591:
### PR summary

* Adds the `FourierFeaturizer`
* Adds a `Pipeline` class resembling that of scikit-learn's, which allows the
  stacking of transformers together.
* Add a class wrapper for `auto_arima`: `AutoARIMA`. This is allows auto-ARIMA
  to be used with pipelines.

__TODO:__ would like to eventually write a decorator in `pmdarima.utils.metaestimators` that will let us more elegantly wrap the `AutoARIMA` functions that delegate down to its fit `ARIMA` class


doc/whats_new.rst
```diff
@@ -28,11 +28,21 @@ v0.8.1) will document the latest features.
 * Deprecate the ``is_stationary`` method in tests of stationarity. This will be removed in
   v1.4.0. Use ``should_diff`` instead.
 
-* Added two new datasets: ``airpassengers`` & ``austres``
+* Add two new datasets: ``airpassengers`` & ``austres``
```
aaronreidsmith: I assume we're trying to keep the "What's new" page in present tense? _Very_ nit-picky, but some of these are past tense (Added the `OCSBTest` of seasonality, as discussed in #88), and some are present tense

tgsmith61591: Yeah we may even want it to be present progressive, like "Version 1.2.0 *adds* <X feature>"

pmdarima/pipeline.py
```diff
@@ -123,7 +122,7 @@ def __len__(self):
     @property
     def named_steps(self):
         # Use Bunch object to improve autocomplete
```
aaronreidsmith: Comment is no longer needed, since we're not using `Bunch`. Not a blocker, but could be confusing for future contributors

PR #131, opened 2019-04-21
[MRG+1] Fix deploy logic on Azure Pipelines

aaronreidsmith:
After doing [a little research](https://stackoverflow.com/a/27154277), and [reading the docs a little bit more](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables), I found that `Build.SourceBranch` would resolve to `refs/tags/<tag>`, and `Build.SourceBranchName` would just resolve to `<tag>`, so the deploy step would fail. I removed the condition that the tag must be on `master`, because as my first link shows, tags and branches are entirely independent.


tgsmith61591:
Great catch. We only tag on master anyways so this works!

PR #135, opened 2019-04-27
[MRG+1] Mac builds on azure

aaronreidsmith:
This PR:

Primarily:
* Adds Mac builds to Azure

Secondarily:
* Makes Azure CI/CD jobs more template-based
* Updated `develop` version to `1.3.0-dev0`
* Removes extraneous quotes from docs
* Brings the following changes from `master`:
  * Copies Travis hotfix to `develop` branch that was previously only on `master`
  * Copies donation button to `develop` that was previously only on `master`
  * Copies `3.7` PyPI classification that was previously only on `master`


tgsmith61591:
Next PR should remove Mac builds from Travis altogether

build_tools/azure/jobs.yml
```diff
@@ -0,0 +1,70 @@
...
+  - template: build_and_deploy.yml
+
+- job: macOS
+
+  pool:
+    vmImage: 'macOS-10.13'  # Also have the option of 10.14, if we want that
+
+  # Only need to set this for macOS
+  variables:
+    PMD_MPL_BACKEND: TkAgg
```
tgsmith61591: Yep, forgot that's how we were [setting it in Travis](https://github.com/tgsmith61591/pmdarima/blob/master/pmdarima/utils/visualization.py#L23)

aaronreidsmith:
> Next PR should remove Mac builds from Travis altogether

Yeah, that is my goal. Just didn't want to rip it out all at once. I will work on that this week!

build_tools/azure/jobs.yml
aaronreidsmith: The only thing I am not doing that Travis does is setting [`PMD_MPL_DEBUG`](https://github.com/tgsmith61591/pmdarima/blob/a133de78ba5bd68da9785b061f519ba28cd514cc/.travis.yml#L23). I don't really know how that is used, is that something we need?

tgsmith61591: We shouldn't for now

aaronreidsmith: Also, it seems we set [`PMDARIMA_MPL_BACKEND`](https://github.com/tgsmith61591/pmdarima/blob/a133de78ba5bd68da9785b061f519ba28cd514cc/.travis.yml#L24) in Travis instead of `PMD_MPL_BACKEND` 🤔

PR #136, opened 2019-04-29
[MRG+1] Remove macOS builds Travis

aaronreidsmith:
This PR:

* Removes macOS builds from Travis
  * I left all the logic that looks for `osx` in case we ever decide to add it back (plus, the Travis files are kind of obfuscated)
* Migrates source builds from Travis to Azure
* Does the README check every time Travis CI runs so we don't have another failure at deploy-time


build_tools/azure/build_and_deploy.yml
```diff
@@ -8,7 +8,16 @@ steps:
   displayName: 'Installing requirements'
 
 - script: python setup.py bdist_wheel
```
tgsmith61591: Can you make this:

```yaml
- script: python setup.py bdist_wheel sdist
```

and it will build both in one go

build_tools/azure/build_and_deploy.yml
```diff
@@ -8,7 +8,16 @@ steps:
...
-  displayName: 'Building artifact'
+  displayName: 'Building wheel'
+
+- script: pip install --pre --no-index --find-links dist/ pmdarima
+  displayName: 'Installing generated wheel'
+
+- script: pytest --showlocals --durations=20 --mpl --mpl-baseline-path=etc/pytest_images --pyargs pmdarima --benchmark-skip
+  displayName: 'Running unit tests'
+
+- script: python setup.py sdist
```
tgsmith61591: ... and then this line can go away

build_tools/azure/build_and_deploy.yml
aaronreidsmith: Implementing that, I get this warning I have never seen before:

```bash
/Users/vsts/hostedtoolcache/Python/3.5.7/x64/lib/python3.5/distutils/dist.py:974: UserWarning: 
`build_src` is being run, this may lead to missing
files in your sdist!  You want to use distutils.sdist
instead of the setuptools version:

    from distutils.command.sdist import sdist
    cmdclass={'sdist': sdist}"

See numpy's setup.py or gh-7131 for details.
  cmd_obj.run()
```

Let me look into that before I merge this

aaronreidsmith: I think it is because in `setup.py`, if we have `bdist_wheel` in the arguments, we set [`we_be_buildin` to `True`](https://github.com/tgsmith61591/pmdarima/blob/a133de78ba5bd68da9785b061f519ba28cd514cc/setup.py#L99). Therefore, we used [`numpy.distutils.core.setup` instead of `setuptools.setup`](https://github.com/tgsmith61591/pmdarima/blob/a133de78ba5bd68da9785b061f519ba28cd514cc/setup.py#L230-L239). Is that the behavior we want?

tgsmith61591: Gotcha. Let's just leave it as is, then. It's not like it's any significant amount of overhead to run the setup.py twice.

build_tools/azure/build_and_deploy.yml
```diff
@@ -7,8 +7,14 @@ steps:
     pip install -r build_tools/azure/requirements.txt
   displayName: 'Installing requirements'
 
-- script: python setup.py bdist_wheel
-  displayName: 'Building artifact'
+- script: python setup.py bdist_wheel sdist
```
tgsmith61591: Yeah, if this is causing a warning (or worse) let's just revert. Sorry 😬

PR #148, opened 2019-06-11
Fixes #147

### LINKED ISSUE #147: pmdarima imports from `sklearn.externals`, which raises a warning
`sklearn.externals.six` and `sklearn.externals.joblib` have been deprecated in the latest version of `sklearn`. Unfortunately, various code in pmdarima imports directly from these deprecated packages. Thus, when I run:

```python
from pmdarima.arima import auto_arima
```

I get a bunch of garbage warnings printed to my stdout:

```
/usr/local/lib/python3.7/site-packages/sklearn/externals/six.py:31: DeprecationWarning: The module is deprecated in version 0.21 and will be removed in version 0.23 since we've dropped support for Python 2.7. Please rely on the official version of six (https://pypi.org/project/six/).
  "(https://pypi.org/project/six/).", DeprecationWarning)
/usr/local/lib/python3.7/site-packages/sklearn/externals/joblib/__init__.py:15: DeprecationWarning: sklearn.externals.joblib is deprecated in 0.21 and will be removed in 0.23. Please import this functionality directly from joblib, which can be installed with: pip install joblib. If this warning is raised when loading pickled models, you may need to re-serialize those models with scikit-learn 0.21+.
  warnings.warn(msg, category=DeprecationWarning)
```

Like the warnings say, it would be better if pmdarima depended upon the actual `six` and `joblib` packages, instead of the ones currently (but not for much longer) bundled with `sklearn`.

telamonian:
All dependencies on `sklearn.externals.{six, joblib}` have been changed to `six` or `joblib`. I also added appropriate entries to `requirements.txt`


telamonian:
There's a `scipy`/`statsmodels` version mismatch that's screwing up the CI tests (`scipy` moved the `factorial` function from one module to another in the latest version).

tgsmith61591:
Thanks for the PR! I changed the target branch to `develop`, since that's where we stage changes until our releases are prepared. We're aware of the statsmodels issue... looks like it's still being worked on from their end:

https://github.com/statsmodels/statsmodels/issues/5747

They have a release candidate out (0.10.0.rc) right now, so the best advice we've been giving up to this point is to downgrade scipy after statsmodels (which is kludgy, and we don't want to have to release a patch that forces a downgrade). That said, @aaronreidsmith we may update our CI scripts to pin scipy at 0.12 until this whole mess is cleared up on statsmodels' end... until then, these CI jobs may be stuck 😬

telamonian:
I tried bumped to the version `statsmodels>=0.10.0rc2` in `requirements.txt` (I committed that right after I posted my previous message, but for some reason it's timestamped before).

The tests are now getting past the issue with `factorial` that I mentioned, but now a handful of specific unit tests are failing for unknown reasons...

telamonian:
@tgsmith61591 What's the command for re-running a specific unit test locally?

tgsmith61591:
Hmm. These are the plotting tests... To run a test locally:

```bash
$ python -m pytest -v pmdarima/path/to/test
```

telamonian:
I did some messing around, and it looks like the unit test failures are all due to changes in `statsmodels==0.10.0rc2`. I guess they made enough changes in the latest version so as to change the behavior of `statsmodels` w.r.t. the expectations of `pmdarima` (or at least its test suite). It looks like `statsmodels` and `scipy` teamed up to make trouble for you guys :)

I reset the `statsmodel` requirement to `>=0.9.0` and pinned `scipy` to `<1.3`, and now all of the CI tests are passing. So, independent of any dependency issues, it looks like the code in this PR is passing and good to go.

More generally, I would strongly recommend that you pull in the change I made to `requirements.txt` (ie `scipy>=1.2, <1.3`) and push a new release to Pypi ASAP, since currently `pmdarima` will be broken in a number of different installation scenarios. For example, `pmdarima` will be broken if you do an upgrade install:

```
pip install -U pmdarima
```

since the resulting version of `scipy` will be too new and the version of `statsmodels` will be too old. Same thing will happen if you just do a regular install in a clean environment (ie a newly created virtual env):

```
pip install pmdarima
```

tgsmith61591:
Actually, since I already merged `develop` into your branch, makes sense to go to `develop` and then do a cherry pick onto master for the release.

PR #155, opened 2019-07-05
[MRG+2] Circle deployments for Linux

tgsmith61591:
Moves PyPi deployments to Circle and backs out all Travis build tools.


Makefile
```diff
@@ -36,6 +36,9 @@ requirements:
 bdist_wheel: requirements
 	$(PYTHON) setup.py bdist_wheel
 
+sdist: requirements
+	$(PYTHON) setup.py sdist
```
aaronreidsmith: We also deploy source on Azure. Do we want to drop one or keep both?

tgsmith61591: I don't think we do:

```yaml
  - script: python setup.py bdist_wheel
    displayName: 'Building artifact'
```

We build the whl file and then deploy all contents of `dist/`, which will only contain the wheel.

tgsmith61591: Oh you're right, I'm looking at `master`, which is behind... 🤦‍♂

Issue #146, opened 2019-06-07
simple_differencing


jeweinberg:
Can you expose the simple_differencing option or allow us to pass kwargs to the ARIMA and SARIMA class instantiation methods into the auto_arima class? This helps in some cases where fitting is prohibitively slow.

tgsmith61591:
Care to take a crack at a PR?

tgsmith61591:
Was hoping we'd get someone else to contribute this feature, but given the amount of time it's been open I merged it in #171. Will be available in 1.3.0 or immediately if building from `master`

PR #171, opened 2019-07-22
[MRG+1] Issue 146

tgsmith61591:
# Description

Adds `**sarimax_kwargs` to the `ARIMA` constructor and `auto_arima` function. This addresses #146. While I was at it, I changed the `plot_diagnostics` unit tests to use `unittest.mock.patch` to make them less brittle. 

## Type of change

# How Has This Been Tested?

- Added a new test that shows that `simple_differencing=True` can fix an `auto_arima`
- Changed `plot_diagnostic` tests to use mocking

# Checklist:


pmdarima/arima/_auto_solvers.py
```diff
@@ -56,6 +57,7 @@ def __init__(self, y, xreg, start_params, trend, method, transparams,
         self.scoring_args = scoring_args
         self.information_criterion = information_criterion
         self.with_intercept = with_intercept
+        self.kwargs = dict() if not kwargs else kwargs
```
aaronreidsmith: Is there a reason to use `dict()` vs `{}`? I know `{}` is faster (https://stackoverflow.com/a/5790954)

pmdarima/arima/auto.py
```diff
@@ -96,6 +97,7 @@ def __init__(self, start_p=2, d=None, start_q=2, max_p=5,
         self.scoring = scoring
         self.scoring_args = scoring_args
         self.with_intercept = with_intercept
+        self.sarimax_kwargs = dict() if not sarimax_kwargs else sarimax_kwargs
```
aaronreidsmith: Same comment regarding `dict()` vs `{}` (just respond to one, just wanted to point both out)

pmdarima/arima/auto.py
```diff
@@ -247,6 +251,8 @@ def auto_arima(y, exogenous=None, start_p=2, d=None, start_q=2, max_p=5,
                                  force_all_finite=True))  # type: np.ndarray
     n_samples = y.shape[0]
 
+    sarimax_kwargs = dict() if not sarimax_kwargs else sarimax_kwargs
```
aaronreidsmith: Same comment (sorry for spam 😂)

pmdarima/arima/_auto_solvers.py
tgsmith61591: `{}` can be a [set literal](https://stackoverflow.com/questions/36674083/why-is-it-possible-to-replace-sometimes-set-with) 😬

aaronreidsmith: The way I understand it, Python will _always_ interpret `{}` as an empty dictionary. So you end up with the following:

```python
example1 = {}  # Empty dict
example2 = {1: 'test'}  # Dict with one item
example3 = set()  # Empty set
example4 = {1, 2, 3}  # Set with 3 items
```
[The official docs even define `{}` as an empty dict.](https://docs.python.org/3.7/library/stdtypes.html#truth-value-testing) The _only_ definition of an empty set is `set()`

Am I off base? I don't think the fraction of a second is a huge deal, but we _do_ get a lot of questions about performance, so I figured I would bring it up!

PR #187, opened 2019-09-07
[MRG+1] Fix deployments, remove old pytest_images dir

tgsmith61591:
This PR fixes Circle deployments and removes the old `pytest_images` directory we no longer use.


build_tools/circle/regression_testing.sh
```diff
@@ -19,6 +22,7 @@ python -c "\
 import joblib
 import pyramid as pm
 
+print('fitting and serialing model with old package')
```
aaronreidsmith: **Typo**: `serialing` -> `serializing`

PR #214, opened 2019-11-07
[MRG+1] Use tags for __version__

### LINKED ISSUE #213: [MAINT] Migrate versioning to tags only
The explicit versioning is a bit of a pain to manage and can fail people's PRs if we aren't on our game. We should switch over to managing the version with a tag.

Rather than set the version explicitly:

```python
__version__ = '1.4.0'
```

We'll do something like:

```python
from pathlib import Path

try:
    version_path = Path(__file__).parent / "VERSION"
    version = version_path.read_text().strip()
except FileNotFoundError:
    version = "0.0.0"
```

Where the `VERSION` file is created by a Make target `make version` only in the presence of tags in CI jobs. That means local or dev versions will all be `0.0.0`.

Things that will need to change:

* [ ] Documentation for dev versions will need to read "dev" as the version rather than 0.0.0, since that would appear clunky and confusing to users
* [ ] `deploy-doc` stage will need to be updated
* [ ] `make` target for Circle (and MacOS jobs?)
* [ ] Windows needs a different approach, since it won't have cmake installed

Anything I'm missing?

aaronreidsmith:
# Description

This PR fixes #213. So far it does the following:

- Adds `make version` target that will:
    - Check to see if it is on Circle or Azure DevOps and create a VERSION file from the tag OR
    - Print that it is not on a CI/CD platform or not a tagged commit
- Adds the `version` target as a dependency on the following targets, so we can't build without creating a VERSION file (assuming we are building using the Makefile):
    - `bdist_wheel`
    - `sdist`
    - `develop`
    - `install`
- Changes the `__version__` variable to be set as described in #213. Tl;dr: it uses the tag version OR `0.0.0`

### TODO:
- [x] Update `deploy-doc` to handle the above changes

## Type of change

# How Has This Been Tested?

for your test configuration

- [X] Tested on Circle and ADO without tags check (to make sure ENV variables worked correctly)
- [X] Tested on Circle and ADO with tags check (to make sure it generates `0.0.0` on CI/CD)
- [X] Tested locally to make sure I get `0.0.0`
- [ ] Tested with an actual tag

# Checklist:


build_tools/azure/build_and_deploy.yml
```diff
@@ -7,7 +7,7 @@ steps:
     pip install -r build_tools/azure/requirements.txt
   displayName: 'Installing requirements'
 
-- script: python setup.py bdist_wheel
+- script: make bdist_wheel
```
tgsmith61591: will this work on windows? is cmake installed?

aaronreidsmith: It works on [Azure Pipelines](https://dev.azure.com/tgsmith61591gh/pmdarima/_build/results?buildId=729). If anyone wants to develop locally on Windows, they would need cmake for it to work...

build_tools/get_tag.py
```diff
@@ -0,0 +1,25 @@
...
+# This file assumes that our tags are always in this format: vX.X.X.
+# In that case, we would only want to write X.X.X
+
+OUT_FILE = Path(__file__).parent / 'VERSION'
+
+# Circle is easy, since they give us the git tag
+if os.getenv('CIRCLECI', False) and os.getenv('GIT_TAG', False):
+    print('Tagged commit on Circle CI. Writing to {0}'.format(OUT_FILE))
+    with open(OUT_FILE, 'w') as file:
+        file.write(os.getenv('GIT_TAG')[1:])
```
tgsmith61591: I see a day in the future where we leave the `v` off the tag either intentionally or accidentally, and then get bitten badly. Just to be sure, let's do something like:

```python
with open(OUT_FILE, 'w') as file:
    tag = os.getenv('GIT_TAG')
    file.write(tag[1:] if tag.startswith('v') else tag)
```

build_tools/azure/build_and_deploy.yml
aaronreidsmith: https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/Vs2017-Server2016-Readme.md#cmake

build_tools/get_tag.py
```diff
@@ -0,0 +1,25 @@
+from pathlib import Path
+import os
+
+# This file assumes that our tags are always in this format: vX.X.X.
+# In that case, we would only want to write X.X.X
+
+OUT_FILE = Path(__file__).parent / 'VERSION'
```
tgsmith61591: I think this needs to be if we want it to be shipped with the package

```python
OUT_FILE = Path(__file__).parent / 'pmdarima' / 'VERSION'
```

aaronreidsmith: Right, but does it need to be shipped with the package? Currently, `pmdarima.__version__` contains all of the version info. That is still happening in this branch, but it just reads from `../VERSION`

build_tools/azure/build_and_deploy.yml
tgsmith61591: Thanks for the sleuthing

build_tools/get_tag.py
aaronreidsmith: Looks like you fixed this!

tgsmith61591: Yes it needs to be shipped since importing `pmdarima` will hit that line in the `__init__` where it tries to load the VERSION. I've included the version in the manifest, so it's bundled with the install.

pmdarima/__init__.py
```diff
@@ -19,10 +20,13 @@
...
 #   X.Y     # Final release
 #
-# Dev branch marker is: 'X.Y.dev' or 'X.Y.devN' where N is an integer.
-# 'X.Y.dev0' is the canonical version of 'X.Y.dev'
-#
-__version__ = "1.4.1.dev0"
+# We only create a VERSION file in CI/CD on tagged commits.
+# For local development, or non-tagged commits, we will use 0.0.0
+try:
+    version_path = Path(__file__).parent / 'VERSION'
```
aaronreidsmith: After 8270ec93bfbbff3057dd7d17d1a85394e4b09ade, this should be `Path(__file__) / 'VERSION'`, right? Or else it will be looking for `../VERSION` and it will ALWAYS get `FileNotFoundError`

tgsmith61591: No, the `parent` attr gives the rel path to the parent dir, which in this case would be `pmdarima`.

```python
# from inside the inner dir:
>>> Path("__init__.py").parent / "VERSION"
'./VERSION'
```

So it's getting the '.' prepended basically.

aaronreidsmith: Ah, then my previous code was wrong. Good catch!

build_tools/get_tag.py
```diff
@@ -0,0 +1,32 @@
+from pathlib import Path
```
aaronreidsmith: After b4f6d96, we can remove this import

PR #221, opened 2019-11-15
[MRG+2] - add StepwiseContext feature

kpsunkara:
# Description

Add Context Manager pattern to allow callers to set runtime context for calls into various functionalities. A new ```StepwiseContext``` content manager is added to enable passing of runtime configuration that offers additional level of control required in situations like, use cases that are either sensitive to duration and/or the number of attempts to find the best fit model. For example: building several hundreds (or even thousands) of models in a reasonable amount of time with a trade off between best fit model and amount of time.

## Type of change

- [x] Documentation change

# How Has This Been Tested?

Added unit tests to cover the new code and code paths and verified they pass the pytest.

- [x] Test validation of arguments to StepwiseContext
- [x] Test calling *auto_arima* with StepwiseContext

# Checklist:

### DIFF
--- a/pmdarima/arima/_auto_solvers.py
+++ b/pmdarima/arima/_auto_solvers.py
@@ -10,7 +10,8 @@ import warnings
 
 from .arima import ARIMA
 from .warnings import ModelFitWarning
-
+from ._context import _context as execution_context
+from datetime import datetime
 
 class _StepwiseFitWrapper(object):
     """The stepwise algorithm fluctuates the more sensitive pieces of the ARIMA
@@ -78,10 +79,14 @@ class _StepwiseFitWrapper(object):
         self.seasonal = seasonal
         self.max_order = max_order
 
+        # execution context passed through the context manager
+        self.exec_context = execution_context()
+
         # other internal start vars
         self.bestfit = None
         self.k = self.start_k = 0
-        self.max_k = 100
+        self.max_k = 100 if self.exec_context.max_steps is None else self.exec_context.max_steps
+        self.max_dur = self.exec_context.max_dur
 
         # results list to store intermittent hashes of orders to determine if
         # we've seen this order before...
@@ -146,6 +151,8 @@ class _StepwiseFitWrapper(object):
                 self.p, self.q, self.P, self.Q = p, q, P, Q
 
     def step_through(self):
+        start_time = datetime.now()
+
         while self.start_k < self.k < self.max_k:
             self.start_k = self.k
 
@@ -182,6 +189,10 @@ class _StepwiseFitWrapper(object):
             self.fit_increment_k_cache_set(self.q < self.max_q and
                                            self.p < self.max_p, q=self.q + 1,
                                            p=self.p + 1)
+
+            # break loop if cumulative execution time exceeds the timeout threshold
+            if self.max_dur and (datetime.now() - start_time).total_seconds() > self.max_dur:
+                break
 
         # return the values
         return self.results_dict.values()
--- a/pmdarima/arima/_auto_solvers.py
+++ b/pmdarima/arima/_auto_solvers.py
@@ -10,7 +10,7 @@ import warnings
 
 from .arima import ARIMA
 from .warnings import ModelFitWarning
-from ._context import _context as execution_context
+from ._context import ContextType, ContextStore
 from datetime import datetime
 
 
@@ -81,7 +81,7 @@ class _StepwiseFitWrapper(object):
         self.max_order = max_order
 
         # execution context passed through the context manager
-        self.exec_context = execution_context()
+        self.exec_context = ContextStore.get_or_empty(ContextType.STEPWISE)
 
         # other internal start vars
         self.bestfit = None
@@ -195,7 +195,14 @@ class _StepwiseFitWrapper(object):
             # break loop if execution time exceeds the timeout threshold
             dur = (datetime.now() - start_time).total_seconds()
             if self.max_dur and dur > self.max_dur:
+                warnings.warn('early termination of stepwise search due to '
+                              'max_dur threshold')
                 break
+
+        # check if the search has ended after max_k tries
+        if self.k > self.max_k:
+            warnings.warn('stepwise search has reached the maximum number '
+                          'of tries to find the best fit model')
 
         # return the values
         return self.results_dict.values()
--- a/pmdarima/arima/_auto_solvers.py
+++ b/pmdarima/arima/_auto_solvers.py
@@ -199,8 +199,9 @@ class _StepwiseFitWrapper(object):
                               'max_dur threshold')
                 break
 
-        # check if the search has ended after max_k tries
-        if self.k > self.max_k:
+        # check if the search has been ended after max_steps
+        if self.exec_context.max_steps is not None \
+                and self.k > self.exec_context.max_steps:
             warnings.warn('stepwise search has reached the maximum number '
                           'of tries to find the best fit model')
 
--- a/pmdarima/arima/_auto_solvers.py
+++ b/pmdarima/arima/_auto_solvers.py
@@ -13,6 +13,7 @@ from .warnings import ModelFitWarning
 from ._context import _context as execution_context
 from datetime import datetime
 
+
 class _StepwiseFitWrapper(object):
     """The stepwise algorithm fluctuates the more sensitive pieces of the ARIMA
     (the seasonal components) first, adjusting towards the direction of the
@@ -85,7 +86,8 @@ class _StepwiseFitWrapper(object):
         # other internal start vars
         self.bestfit = None
         self.k = self.start_k = 0
-        self.max_k = 100 if self.exec_context.max_steps is None else self.exec_context.max_steps
+        self.max_k = 100 if self.exec_context.max_steps is None \
+            else self.exec_context.max_steps
         self.max_dur = self.exec_context.max_dur
 
         # results list to store intermittent hashes of orders to determine if
@@ -190,8 +192,9 @@ class _StepwiseFitWrapper(object):
                                            self.p < self.max_p, q=self.q + 1,
                                            p=self.p + 1)
 
-            # break loop if cumulative execution time exceeds the timeout threshold
-            if self.max_dur and (datetime.now() - start_time).total_seconds() > self.max_dur:
+            # break loop if execution time exceeds the timeout threshold
+            dur = (datetime.now() - start_time).total_seconds()
+            if self.max_dur and dur > self.max_dur:
                 break
 
         # return the values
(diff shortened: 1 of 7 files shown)


pmdarima/arima/auto.py
```diff
@@ -173,6 +174,44 @@ def update(self, y, exogenous=None, maxiter=None, **kwargs):
...
+    max_dur : int, optional (default=None)
+        The maximum duration in seconds to try to find a best fit.
+        When the cumulative fit duration exceeds this number, the
+        stepwiese process will stop and the best fit model at that
+        time will be returned. Please note that this is a soft limit.
+    """
+
+    def __init__(self, max_steps=100, max_dur=None):
+        if max_steps is None or not 0 < max_steps <= 100:
+            raise ValueError('max_steps should be between 1 and 100')
```
tgsmith61591: I wonder whether it makes sense to allow the user to specify more than 100? The default currently is 100 in the stepwise wrapper, but I like the pythonic contract of `I'm handing you the gun but trusting you not to shoot yourself in the foot`. If the user wants 1000, should we prevent that?

pmdarima/arima/tests/test_arima.py
```diff
@@ -1065,3 +1065,26 @@ def test_issue_191():
...
+    ])
+def test_stepwise_context_args(max_steps, max_dur):
+    with pytest.raises(ValueError):
+        StepwiseContext(max_steps=max_steps, max_dur=max_dur)
+
+
+# test auto_arima stepwise run with  StepwiseContext
+def test_auto_arima_with_stepwise_context():
+    samp = lynx[:8]
+    with StepwiseContext(max_steps=5, max_dur=30):
```
tgsmith61591: Any way in this test to assert that only 5 steps are taken?

pmdarima/arima/_context.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+#
+# re-entrant, reusable context manager to store execution context
+#
+
+import threading
+
+# thread local value to store the context info
+_ctx = threading.local()
```
tgsmith61591: I have a few concerns about using a single global thread local to store all the attributes:

1. A nested context would break (I don't know who would ever do this, but if they did we'd have a problem..):

  ```python
  with StepwiseContext(max_steps=5, max_dur=30) as ctx1:
      with StepwiseContext(max_steps=10, max_dur=15) as ctx2:
          # I would expect ctx1.max_steps to be 5, but I think it would be 10 by the time auto_arima was called
  ```

2. The use of multiple `StepwiseContext`s in a script could produce confusing results:

  ```python
  with StepwiseContext(max_dur=30) as ctx1:
      fit = auto_arima(y, **kwargs)

  with StepwiseContext(max_steps=5) as ctx2:
      # now I think ctx2 would confusingly inherit the max_dur from the first one
  ```

I understand that we need a global so that the stepwise wrapper can acquire it without it being passed... what if rather than a single thread local, the global were a `collections.deque` like:

```python
# _context.py
import collections

_ctx = collections.deque()

class _context():
    def __init__(self, **kwargs):
        for key in kwargs:
            setattr(self, key, kwargs[key])
        _ctx.append(self)  # this can be "popped" via first-in-last-out
```

.. so in the stepwise fit wrapper the context was called more like:

  ```python
  # _auto_solvers.py
  from ._context import _ctx

  ...
  # inside __init__
  try:
      self.exec_context = _ctx.pop()
  except IndexError:
      self.exec_context = None  # not in a context manager
  ```

Thoughts?

kpsunkara: I think that's a great idea!

Perhaps we may want to leave it as a thread specific instance, but store context instances in a `deque`. An context is added to deque on context manager `__enter__` and removed on `__exit__`, so we can support different usage patterns using `ExitStack`, etc.

To generalize the approach to support other context managers in future, we can wrap access to the thread local into a set of convenience methods (with other supporting fixtures) to something like...

```python
import threading
from abc import ABC, abstractmethod
from enum import Enum
import collections

# thread local value to store the context info
_ctx = threading.local()
_ctx.store = {}

__all__ = ['AbstractContext', 'ContextStore', 'ContextType']


class ContextType(Enum):
    """Context Type Enumeration

    An enumeration of Context Types known to ``ContextStore``

    """
    EMPTY = 0
    STEPWISE = 1


class AbstractContext(ABC):
    """An abstract context manager to store execution context.

    A generic, re-entrant, reusable context manager to store
    execution context in a threading.local instance. Has helper
    methods to iterate over the context info and provide a
    string representation of the context info.

    """
    def __init__(self, **kwargs):
        self.props = kwargs if kwargs else {}

    def __enter__(self):
        ContextStore._add_context(self)

    def __exit__(self, exc_type, exc_val, exc_tb):
        ContextStore._remove_context(self)

    def __getattr__(self, item):
        return self.props[item] if item in self.props else None

    def __contains__(self, item):
        return item in self.props

    def __getitem__(self, item):
        return self.props[item] if item in self.props else None

    def __iter__(self):
        for key in self.props:
            yield key

    def __repr__(self):
        return self.props.__repr__()

    @abstractmethod
    def get_type(self):
        return None


class _emptyContext(AbstractContext):

    def __init__(self):
        super(_emptyContext, self).__init__()

    def get_type(self):
        return ContextType.EMPTY

class ContextStore:
    """A class to wrap access to threading.local() context store

    This class hosts static methods to wrap access to and
    encapsulate the threading.local() instance
    """
    @staticmethod
    def get_context(context_type):
        """Returns most recently added instance of given Context Type

        :param context_type: Context Type to retrieve from the store
        :return: an instance of AbstractContext subclass or None
        """
        if not isinstance(context_type, ContextType):
            raise ValueError('context_type must be an instance of ContextType')

        if context_type in _ctx.store and len(_ctx.store[context_type]) > 0:
            return _ctx.store[context_type][-1]

    @staticmethod
    def get_or_default(context_type, default):
        """Returns most recent instance of given Context Type or default

        :param context_type: Context Type to retrieve from the store
        :param default: value to return in case given context does not exist
        :return: an instance of AbstractContext subclass or default
        """
        ctx = ContextStore.get_context(context_type)
        return ctx if ctx else default

    @staticmethod
    def get_or_empty(context_type):
        """Returns recent instance of given Context Type or an empty context

        :param context_type: Context Type to retrieve from the store
        :return: an instance of AbstractContext subclass
        """
        return ContextStore.get_or_default(context_type, _emptyContext())

    @staticmethod
    def _add_context(ctx):
        """Add given instance of AbstractContext subclass to context store

        This private member is only used by ``AbstractContext``

        :param ctx:
        :return: None
        """
        if not isinstance(ctx, AbstractContext):
            raise ValueError('ctx must be be an instance of AbstractContext')

        context_type = ctx.get_type()
        if context_type not in _ctx.store:
            _ctx.store[context_type] = collections.deque()

        _ctx.store[context_type].append(ctx)

    @staticmethod
    def _remove_context(ctx):
        """Removes the most recently added context of given Context Type

        This private member is only used by ``AbstractContext``
        :param ctx:
        :return: None
        """
        if not isinstance(ctx, AbstractContext):
            raise ValueError('ctx must be be an instance of AbstractContext')

        context_type = ctx.get_type()

        if context_type not in _ctx.store or len(_ctx.store[context_type]) == 0:
            return

        _ctx.store[context_type].pop()

```

Does it make sense?

tgsmith61591: Looks cool. Would love to see it in action :)

pmdarima/arima/tests/test_arima.py
```diff
@@ -1065,3 +1066,64 @@ def test_issue_191():
...
+# test param validation of ContextStore's add, get and remove members
+def test_add_get_remove_context_args():
+    with(pytest.raises(ValueError)):
+        ContextStore._add_context(None)
+
+    with(pytest.raises(ValueError)):
+        ContextStore._remove_context(None)
+
+    with(pytest.raises(ValueError)):
+        ContextStore.get_context(None)
```
aaronreidsmith: Is there a reason the `with` keyword is being used as a function in these tests?

I'd also probably like to see this syntax, as opposed to nested context managers (may have to wrap due to line length):

```python
with StepwiseContext(**ctx1_data) as ctx1, StepwiseContext(**ctx2_data) as ctx2:
    ...
```

pmdarima/arima/_auto_solvers.py
```diff
@@ -183,6 +192,18 @@ def step_through(self):
...
+            dur = (datetime.now() - start_time).total_seconds()
+            if self.max_dur and dur > self.max_dur:
+                warnings.warn('early termination of stepwise search due to '
+                              'max_dur threshold')
+                break
+
+        # check if the search has ended after max_k tries
+        if self.k > self.max_k:
+            warnings.warn('stepwise search has reached the maximum number '
+                          'of tries to find the best fit model')
```
tgsmith61591: Should this `break`?

pmdarima/arima/tests/test_arima.py
```diff
@@ -1065,3 +1066,64 @@ def test_issue_191():
...
+        with(ctx2):
+            effective_ctx_data = ContextStore.get_or_empty(ContextType.STEPWISE)
+            expected_ctx_data = ctx1_data.copy()
+            expected_ctx_data.update(ctx2_data)
+
+            assert all(effective_ctx_data[key] == expected_ctx_data[key]
+                       for key in expected_ctx_data.keys())
+
+            assert all(effective_ctx_data[key] == expected_ctx_data[key]
+                       for key in effective_ctx_data.keys())
```
tgsmith61591: Is this asserting the same thing as above?

kpsunkara: I'm trying to assert that: _effective &bigcap; expected_ == _expected &bigcap; effective_

pmdarima/arima/_auto_solvers.py
kpsunkara: at this point, the control is already outside the ``while`` loop. The warning helps us to assert that stepwise ran for max_steps times.

pmdarima/arima/tests/test_arima.py
kpsunkara: I agree.

pmdarima/arima/_auto_solvers.py
kpsunkara: I'll be making a change to raise this warning only when ``max_steps`` threshold was supplied through the ``StepwiseContext`` and ``self.k > max_steps``. Otherwise, this warning always pops up!

tgsmith61591:
Merging up so other PR can rebase off this. Thanks for the awesome change @kpsunkara! If there are any other minor additions you wanted to make, we can take them in another PR

kpsunkara:
Thank you, @tgsmith61591 and @aaronreidsmith for the awesome feedback! Happy to contribute!!

PR #223, opened 2019-11-18
[MRG+1] Migrate downloads badges

aaronreidsmith:
# Description

This PR moves the script that updates out downloads badges from running on [Zapier](https://zapier.com) to running on Github Actions. Zapier has a 1-second timeout on its Python scripts and fails about 25% of the time. This version will still use [Zapier's data store](https://store.zapier.com), because it is incredibly convenient, but it will not run on Zapier

## Type of change

# How Has This Been Tested?

I have run this script locally and it works fine. I have not been able to test on GitHub actions. Worst case, it fails and our badges don't update. Since they are in the thousands and millions, it will be negligible anyway.

# Checklist:


etc/downloads_badges.py
```diff
@@ -1,19 +1,12 @@
...
 import math
 import os
 import requests
 
 
-# This function is duplicated from `zapier_total_downloads`. This is intentional,
-# because Zapier only has access to one script at a time, so we cannot import
 def millify(n):
-    """Abreviate a number to nearest thousand, million, etc.
+    """Abbreviate a number to nearest thousand, million, etc.
```
tgsmith61591: 👏

aaronreidsmith: Can't believe that got approved the first time, smh

PR #230, opened 2019-11-22
[MRG+1] Polyroot test

### LINKED ISSUE #208: stationarity and invertibility checks in auto-arima
does pmdarima auto-arima check for invertibility and stationarity of the selected model? does it also check for almost-non-invertibility and almost-non-stationarity like in the R implementation?

tgsmith61591:
# Description

What's changing?
* This fixes #208 by adding a poly root test of near non-invertibility when `stepwise=True`
* Rewrite the stepwise solver to be more readable and less stateful. Also updates the rules to be in parity with `forecast::auto.arima`'s newer logic.
* `max_order` is now 5 as a default, rather than 10, and no longer is forced as a constraint when `stepwise=True`

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Documentation change

# How Has This Been Tested?

Existing tests pass, as well as new tests.

# Checklist:


pmdarima/arima/_auto_solvers.py
```diff
@@ -105,55 +161,77 @@ def is_new_better(self, new_model):
...
+        # fit a baseline p, d, q model
+        self._do_fit((p, d, q), (P, D, Q, m))
+
+        # null model with possible constant
+        if self._do_fit((0, d, 0), (0, D, 0, m)):
+            p = q = P = Q = 0
+
+        # A basic AR model
+        if max_p > 0 or max_P > 0:
+            _p = int(max_p > 0)
```
aaronreidsmith: I feel like these are a little more readable like this:

```python
_p = 1 if max_p > 0 else 0
```

It's also a little faster:

```python
>>> import timeit
>>> timeit.timeit("max_p = 1; _p = int(max_p > 0)", number=100000)
0.018097285999999713
>>> timeit.timeit("max_p = 1; _p = 1 if max_p > 0 else 0", number=100000)
0.005227653000019927
```

Thoughts?

tgsmith61591: Pretty amazing that there is such a time differential. Even as minute as it really is... will change

aaronreidsmith: I thought it might be because it didn't have to evaluate the second half of the expression, but it is just as fast using `max_p = 0`:

```python
>>> import timeit
>>> timeit.timeit("max_p = 0; _p = int(max_p > 0)", number=100000)
0.017676304000019627
>>> timeit.timeit("max_p = 0; _p = 1 if max_p > 0 else 0", number=100000)
0.004342629000007037
```

PR #242, opened 2019-12-01
[MRG+2] Seasonal decomposition

### LINKED ISSUE #190: add seasonal decomposition
**Is your feature request related to a problem? Please describe.**
provide a method to produce the seasonal decomposition

**Describe the solution you'd like**
a visual showing the trend, seasonality, and residual, similar to how its done in R.  Ensure visuals can be modified using Matplotlib, Seaborn or other. 

**Describe alternatives you've considered**
currently we have to use statsmodels to run separate script to produce the decomposition

charlesdrotar:
# Description

Fix for mirroring R's `decompose` function as per [issue#190](https://github.com/alkaline-ml/pmdarima/issues/190)

Fixes #190 

## Type of change

# How Has This Been Tested?

- [x] Verified the behavior with R's implementation as per [this blog](https://anomaly.io/seasonal-trend-decomposition-in-r/index.html) by recreating the behavior as 2 new notebooks under the `examples` folder.
- [x] Unittests

# Checklist:


pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,82 @@
...
+    m = np.reshape(detrend, (int(detrend.shape[0] / f), f))
+    seasonal = np.nanmean(m, axis=0).tolist()
+    seasonal = np.array(seasonal[half_f:] + seasonal[:half_f])
+    temp = seasonal
+    for i in range(m.shape[0]):
+        seasonal = np.concatenate((seasonal, temp))
+
+    # We buffer the trend component so that it is the same length as the other
+    # outputs. This counters the effects of losing data by the convolution/sma
+    buffer = []
```
aaronreidsmith: I think it is a little faster to write it like this: `buffer = [np.nan] * half_f`

```python
>>> import timeit
>>> timeit.timeit("""
... import numpy as np
... half_f = 100
... buffer = []
... for i in range(half_f):
...     buffer.append(np.nan)
... """, number=100000)
0.9611893529999804
```

```python
>>> import timeit
>>> timeit.timeit("""
... import numpy as np
... half_f = 100
... buffer = [np.nan] * half_f
... """, number=100000)
0.06795452099999011
```

charlesdrotar: Nice! That’s a heck of a lot faster!

aaronreidsmith:
[Issue : Seasonal Decomposition PR](https://trello.com/c/lPjvCAXd/47-issue-seasonal-decomposition-pr)

pmdarima/arima/seasonality.py
charlesdrotar: Done!

pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,124 @@
 
 __all__ = [
     'CHTest',
+    'decompose',
     'OCSBTest'
 ]
 
 
+def decompose(x, type, f, filter=None):
+    """
```
tgsmith61591: Can you please write up a docstr :)

pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,124 @@
...
+
+    Parameters
+    ----------
+    x : np.array
```
tgsmith61591: Maybe `x : np.array, shape=(n_samples,)`?

pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,124 @@
...
+        The time series of which the trend, seasonal, and noise/random
+        components will be extracted.
+
+    type: str
```
tgsmith61591: Do you think it's worth starting with a default? Or is that something we should not make assumptions about? Also, how do you feel about renaming this to `typ` since `type` is a reserved word in Python?

pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,124 @@
...
+        The type of decomposition that will be performed - 'multiplicative' or
+        'additive'. We would use 'multiplicative' generally when we see an
+        increasing trend. We use 'additive' when the trend is relatively
+        stable over time.
+
+    f: int
```
tgsmith61591: Can we rename this `m`? We use `m` across the rest of the package to denote seasonal frequency.

pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,124 @@
...
+        The frequency in terms of number of observations. This behaves
+        similarly to R's frequency for a time series (ts).
+
+    filter: np.array
```
tgsmith61591: Can you make this `filter: np.array, optional (default=None)`? Also, `filter` is another reserved word in Python. Would it make sense to rename to `filt` (or whatever you like)?

pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,124 @@
...
+           https://www.rdocumentation.org/packages/stats/versions/3.6.1/topics/decompose
+
+
+    """
+
+    multiplicative = "multiplicative"
+    additive = "additive"
+
+    # Helper function to stay consistent and concise based on 'type'
+    def _decomposer_helper(a, b, type):
```
tgsmith61591: You shouldn't need to pass `type` (or `typ` if you rename it) since it's in a closure

pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,124 @@
...
+        if type == multiplicative:
+            return a / b
+        else:
+            return a - b
+
+    # Since R's ts class has a frequency as input I think this it acceptable
+    # to ask the user for the frequency.
+    try:
+        assert int(f) > 1 and isinstance(f, int)
```
tgsmith61591: If you reverse the order of this assertion, you won't need to cast `f` (`m` after it's renamed) to an int

pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,124 @@
...
+        err_msg = "'type' can only take values '{}' or '{}'"
+        raise ValueError(err_msg.format(multiplicative, additive))
+
+    # There needs to be at least 2 periods. This is due to the behavior of
+    # convolutions and how they behave with respect to losing endpoints
+    if (x.shape[0] / f) < 2:
+        raise ValueError("time series has no or less than 2 periods")
+
+    # Take half of f for the convolution / sma process.
+    half_f = int(f / 2.)
```
tgsmith61591: Unimportant nitpick, since you know `m` (`f`) is an int now, you could just do `f // 2`

pmdarima/arima/seasonality.py
```diff
@@ -26,10 +28,124 @@
...
+    m = np.reshape(detrend, (int(detrend.shape[0] / f), f))
+    seasonal = np.nanmean(m, axis=0).tolist()
+    seasonal = np.array(seasonal[half_f:] + seasonal[:half_f])
+    temp = seasonal
+    for i in range(m.shape[0]):
+        seasonal = np.concatenate((seasonal, temp))
+
+    # We buffer the trend component so that it is the same length as the other
+    # outputs. This counters the effects of losing data by the convolution/sma
+    buffer = [np.nan] * half_f
```
tgsmith61591: Another nice way to do this (doesn't matter, because with a one-off like this you won't get a boost. but since @aaronreidsmith commented, I will as well) is simply `np.ones(half_f) * np.nan`

pmdarima/utils/visualization.py
```diff
@@ -58,6 +59,49 @@ def _show_or_return(obj, show):
...
+
+    """
+
+    _err_for_no_mpl()
+
+    fig, axes = mpl.subplots(4, 1, sharex=True, **figure_kwargs)
+
+    y_labels = ['data', 'trend', 'seasonal', 'random']
+
+    for ax in axes.flat:
```
tgsmith61591: This is a neat trick, didn't know you could do that.

pmdarima/arima/seasonality.py
charlesdrotar: I would not put a default. I would force the user to be mindful about what is happening. I believe R does the same thing. They do filter the acceptable values to `multiplicative` or `additive`. See R behavior [here](https://www.rdocumentation.org/packages/stats/versions/3.6.1/topics/decompose).

pmdarima/arima/seasonality.py
charlesdrotar: So I used `f` because that is what was used internally within the R source code for the `decompose` function. I think we can put `m` and just put a little note that it sometimes denoted as `f` within the R code.

pmdarima/arima/seasonality.py
charlesdrotar: I can evaluate to see if it is gives any type of boost. If so, I can definitely make the change.

pmdarima/utils/visualization.py
charlesdrotar: Yup! So it is technically a `numpy` array under the hood so you can do that.

pmdarima/arima/seasonality.py
charlesdrotar: D'oh

pmdarima/arima/seasonality.py
charlesdrotar: Definitely will rename `type` 👍

pmdarima/arima/seasonality.py
tgsmith61591: Yeah, but we're not 1:1 with R. I'd rather be consistent within our package than across different software.

pmdarima/arima/seasonality.py
charlesdrotar: It seems like the preliminary tests were inconclusive so I will just leave as is.

pmdarima/arima/seasonality.py
charlesdrotar: On second thought...I will follow [PEP8](https://pep8.org/#descriptive-naming-styles) and just add a single trailing underscore for python reserved keywords.

pmdarima/arima/seasonality.py
charlesdrotar: I will follow PEP8 and just add a single trailing underscore for python reserved keywords. :). See above reply to link to PEP8 docs.

pmdarima/arima/seasonality.py
charlesdrotar: Done!

pmdarima/arima/seasonality.py
charlesdrotar: Done!

pmdarima/arima/seasonality.py
charlesdrotar: Done!

pmdarima/arima/seasonality.py
charlesdrotar: Done!

pmdarima/arima/seasonality.py
charlesdrotar: Done!

pmdarima/arima/seasonality.py
charlesdrotar: Done!

pmdarima/arima/seasonality.py
charlesdrotar: Done!

pmdarima/arima/seasonality.py
charlesdrotar: Done!

PR #246, opened 2019-12-03
[MRG+1] General cleanup

tgsmith61591:
General clean up including, but not limited to:

* Adding link in API ref for `pm.decompose`
* Add `decompose` to top-level API so you can just do `pm.decompose`
* Clean out some old, pre-Python3 imports
* Add an `__init__.py` to `model_selection/tests` which *may* have contributed to breaking build-from-source builds

### DIFF
--- a/pmdarima/arima/seasonality.py
+++ b/pmdarima/arima/seasonality.py
@@ -2,9 +2,7 @@
 #
 # Author: Taylor Smith <taylor.smith@alkaline-ml.com>
 #
-# Tests for seasonal differencing terms
-
-from __future__ import absolute_import, division
+# Tests for seasonal differencing terms, and seasonal decomposition
 
 from collections import namedtuple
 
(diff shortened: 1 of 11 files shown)


pmdarima/arima/seasonality.py
```diff
@@ -59,28 +57,27 @@ def decompose(x, type_, m, filter_=None):
...
-        where `x` is the input signal, `trend` is the overall trend, `seasonal`
-        is the seasonal component, and `random` is the noisy component. The
-        input signal `x` can be mostly reconstructed by the other three
-        components with a number of points missing equal to `f`.
+        A named tuple with ``x``, ``trend``, ``seasonal``, and ``random``
+        components where ``x`` is the input signal, ``trend`` is the overall
+        trend, ``seasonal`` is the seasonal component, and `random` is the
+        noisy component. The input signal ``x`` can be mostly reconstructed by
+        the other three components with a number of points missing equal to
+        ``m``.
```
tgsmith61591: RST needs double '``' for properly rendering code in the doc. Hence this change.

pmdarima/setup.py
```diff
@@ -4,9 +4,8 @@
 #
 # Setup for submodules of pmdarima
 
-from __future__ import absolute_import
-
 import os
+import sys
```
aaronreidsmith: Don't think this is used except for the `TODO`

tgsmith61591: True 👀

Issue #252, opened 2019-12-05
Sklearn Deprecation Warnings for versions 0.23 and 0.24


charlesdrotar:
**Describe the bug**

After the most recent upgrade of sklearn it has introduced the following deprecation warnings. They really don't need to be resolved until the version of `sklearn` is upgraded to where the changes are breaking. 

- 0.23
  - [x] `check_is_fitted`
    ```
    /Users/runner/hostedtoolcache/Python/3.6.9/x64/lib/python3.6/site-packages/sklearn/utils/validation.py:933: FutureWarning: Passing attributes to check_is_fitted is deprecated and will be removed in 0.23.
    ```

- 0.24
  - [ ] `safe_indexing`
     ```
     /Users/runner/hostedtoolcache/Python/3.6.9/x64/lib/python3.6/site-packages/sklearn/utils/deprecation.py:87: FutureWarning: Function safe_indexing is deprecated; 
     safe_indexing is deprecated in version 0.22 and will be removed in version 0.24.
     ``` 

  - [x] `sklearn.metrics.regression` 
     ```
     FutureWarning: The sklearn.metrics.regression module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.metrics. Anything that cannot be imported from sklearn.metrics is now part of the private API.
     ```

**To Reproduce**  

N/A

**Versions**
`pmdarima` = `1.4.0`
`sklearn` < `0.23`

**Expected behavior**
No warnings triggered.

**Actual behavior**
Warnings are being triggered.

**Additional context**
Keeping this here more for housekeeping so that as `0.23` and `0.24` are released we can proactively anticipate these potentially breaking changes.

aaronreidsmith:
Added `sklearn.metrics.regression` to 0.24 as well

tgsmith61591:
I'm pretty sure this was fixed. We can reopen if we find any more.

PR #255, opened 2019-12-08
[MRG+2] Sklearn deprecation fixes

charlesdrotar:
# Description

Addresses the follow deprecation warnings from [issue#252](https://github.com/alkaline-ml/pmdarima/issues/252):

`0.23`

- [x] `check_is_fitted`
  - `/Users/runner/hostedtoolcache/Python/3.6.9/x64/lib/python3.6/site-packages/sklearn/utils/validation.py:933: FutureWarning: Passing attributes to check_is_fitted is deprecated and will be removed in 0.23.`

`0.24`

- [x] sklearn.metrics.regression
  - `FutureWarning: The sklearn.metrics.regression module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.metrics. Anything that cannot be imported from sklearn.metrics is now part of the private API.`

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)expected)

# How Has This Been Tested?

- [x] Validate that warnings dropped
- [x] Existing tests passed

# Checklist:

### DIFF
--- a/pmdarima/compat/sklearn.py
+++ b/pmdarima/compat/sklearn.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Charles Drotar <drotarcharles@gmail.com>
+#
+# Patch backend for sklearn
+
+from __future__ import absolute_import
+
+import sklearn
+from sklearn.utils.validation import check_is_fitted
+
+__all__ = [
+    'get_compatible_check_is_fitted'
+]
+
+
+def get_compatible_check_is_fitted(estimator, attributes=None):
+    """Make the ``check_is_fitted`` function compatible across the
+    sklearn 0.21 to 0.22+ boundary.
+
+
+    Parameters
+    ----------
+    estimator : estimator instance,
+        The estimator that will be checked to see if it is fitted.
+
+    attributes : str, optional (default=None)
+        The attributes to be passed into the ``check_is_fitted`` function.
+
+    """
+
+    # Split the version string from sklearn and use that determine how to use
+    # the `sklearn.utils.validation import check_is_fitted` function
+    semantic_version_split = [int(x) for x in sklearn.__version__.split(".")]
+    major_version = semantic_version_split[0]
+    minor_version = semantic_version_split[1]
+
+    # Determine if we should require a value for the attribute field
+    is_no_attribute_needed = False
+    if major_version == 0 and minor_version > 21:
+        is_no_attribute_needed = True
+
+    # If we need to pass in `attributes` but it is not a str type raise error.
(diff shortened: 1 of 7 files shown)


pmdarima/compat/sklearn.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Charles Drotar <drotarcharles@gmail.com>
+#
+# Patch backend for sklearn
+
+from __future__ import absolute_import
```
tgsmith61591: We prob don't need this anymore because now we're only python 3+

pmdarima/compat/sklearn.py
```diff
@@ -0,0 +1,53 @@
...
+    attributes : str, optional (default=None)
+        The attributes to be passed into the ``check_is_fitted`` function.
+
+    """
+
+    # Split the version string from sklearn and use that determine how to use
+    # the `sklearn.utils.validation import check_is_fitted` function
+    semantic_version_split = [int(x) for x in sklearn.__version__.split(".")]
+    major_version = semantic_version_split[0]
+    minor_version = semantic_version_split[1]
```
tgsmith61591: You could probably do this a bit more elegantly like:

```python
>>> import pkg_resources
>>> pkg_resources.parse_version(sklearn.__version__)
<Version('0.21.2')>
```

It's nice because it allows comparisons like

```python
if parse_version(sklearn.__version__) >= parse_version('0.22.0')
```

tgsmith61591: This would also prevent breakage from anyone running a local version of scikit that may be versioned, e.g., `0.24.0.dev0`

pmdarima/compat/sklearn.py
```diff
@@ -0,0 +1,53 @@
...
+        is_no_attribute_needed = True
+
+    # If we need to pass in `attributes` but it is not a str type raise error.
+    if not is_no_attribute_needed and not isinstance(attributes, str):
+        raise TypeError("Parameter `attributes` can only accept str type")
+
+    # If we don't need an attribute (i.e. sklearn.__version__ >= 0.22)
+    # We ignore `attributes` to bypass warning
+    if is_no_attribute_needed:
+        return check_is_fitted(estimator=estimator)
```
tgsmith61591: I think we're going to have to inspect what changes they're making to the core function to determine whether it's been fitted and make sure we've implemented that in our estimator, accordingly.

pmdarima/compat/sklearn.py
charlesdrotar: Ahhh good call! Will do!

PR #174, opened 2019-08-06
[WIP] Conda deploy

aaronreidsmith:
# Description

This PR mainly adds conda builds for the following OSes/versions:

|OS|Python|
|---|---|
|Windows (x64)|3.6|
|Windows (x64)|3.7|
|macOS|3.6|
|macOS|3.7|
|Linux|3.6|
|Linux|3.7|

#### Why just those versions?
The main reason we can't use 3.5 is because both [SciPy](https://anaconda.org/anaconda/scipy/files) _and_ [Cython](https://anaconda.org/anaconda/cython/files) (and maybe others) don't build conda distributions for Python versions <3.6.

#### Why just x64?
The main reason we can't use x86 machines is because even though we can mimic 32-bit machines for building our wheels, conda requires actual 32-bit hardware, which we don't have access to.

#### In-depth description
- The root of a conda recipe is the `meta.yaml` file, which essentially contains everything from requirements to documentation
  - Since our requirements are constantly changing, this file is rendered at build time from our `requirements.txt`file
- On Azure DevOps our conda distribution is part of our main jobs, so rather than re-invent the wheel (pun intended), we build the conda distribution _from_ our wheel files
- On Circle CI, we build from source
- Our Azure DevOps builds now live in one parameterized job, and we just have to supply variables such as `python.version` or `build_conda`

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation change

# How Has This Been Tested?

- [X] Run our test suite using conda on all OS/Version combinations above (via `conda-build`)

# Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] New and existing unit tests pass locally with my changes

PR #263, opened 2019-12-11
[MRG+2] First-party Conda Distributions

### LINKED ISSUE #173: unable to coda install pmdarima
**Describe the bug**

**To Reproduce**  
typed the following in terminal - 
conda install -c saravji pmdarima

```python
(Pk) prasads-imac:Pk prasadkamath$ conda install -c saravji pmdarima
Collecting package metadata (repodata.json): done
Solving environment: failed
Initial quick solve with frozen env failed.  Unfreezing env and trying again.
Solving environment: 
```

**Versions**

Darwin-18.6.0-x86_64-i386-64bit
Python 3.5.6 |Anaconda, Inc.| (default, Aug 26 2018, 16:30:03) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
Traceback (most recent call last):
  File "/Users/prasadkamath/.conda/envs/Pk/lib/python3.5/site-packages/IPython/core/interactiveshell.py", line 2961, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-4-df0eca425da7>", line 3, in <module>
    import pmdarima; print("pmdarima", pmdarima.__version__)
  File "/Applications/PyCharm CE.app/Contents/helpers/pydev/_pydev_bundle/pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
ImportError: No module named 'pmdarima'

**Expected behavior**

**Actual behavior**

**Additional context**

aaronreidsmith:
# Description

This PR supersedes #174 and fixes #173. There is a lot going on, so let's break it down:

1. Migrates all jobs previously hosted on Azure Pipelines to Github Actions
    - This was done because Azure Pipelines has better support for conda, and we want to maximize our parallelism
2. Adds first-party conda distributions for all platforms **that our dependencies support**
    - Our dependencies only support 64-bit Python 3.6+ builds ([Exhibit A](https://anaconda.org/anaconda/numpy/files), [Exhibit B](https://anaconda.org/anaconda/cython/files))

#### Supported Versions
|OS|Python|
|---|---|
|Windows (x64)|3.6|
|Windows (x64)|3.7|
|macOS|3.6|
|macOS|3.7|
|Linux|3.6|
|Linux|3.7|

#### TODO:
- [x] Add `TWINE_USERNAME` and `TWINE_PASSWORD` secrets to github
- [x] Add a deploy step for conda

## Type of change

# How Has This Been Tested?

- [X] All existing tests pass on all CI/CD platforms
- [X] Tests are run on the conda version of the package, as well as an installation test

# Checklist:


aaronreidsmith:
I still need to add the badge for GitHub Actions. The name of the workflow will appear in the badge. Would y'all like to change the name, or do you like it?

Github Badge

aaronreidsmith:
The 9 expected checks will not run, because they no longer exist. It's like it is trying to run the old YAML checks AND the new YAML checks 🤔

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,104 @@
...
+    branches:
+      - '*'
+
+jobs:
+  build-and-deploy:
+    name: Build and Deploy
+    strategy:
+      matrix:
+        os: [windows-latest, macos-latest]
+        python-version: ['3.5', '3.6', '3.7']  # We build 3.5 on Azure Pipelines
```
tgsmith61591: You mean we build 3.8 on Azure pipelines?

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,104 @@
...
+      # https://social.msdn.microsoft.com/Forums/vstudio/en-US/06675e6d-9a0b-46f5-8de0-10237fab1ece/link-error-lnk1158?forum=vcgeneral
+      - name: Fix Windows
+        if: matrix.os == 'windows-latest'
+        run: |
+          Copy-Item "C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\x86\rc.exe" -Destination "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN"
+          Copy-Item "C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\x86\rcdll.dll" -Destination "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN"
+        shell: powershell
+
+      - name: Setting up Python
+        uses: actions/setup-python@v1.1.1
```
tgsmith61591: What is this doing?

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,104 @@
...
+        shell: bash
+
+      - name: Installing generated wheel
+        run: pip install --pre --no-index --find-links dist/ pmdarima
+        shell: bash
+
+      - name: Running unit tests
+        run: |
+          if [ "${{ matrix.os }}" == "macos-latest" ]; then
+            export PMD_MPL_BACKEND=TkAgg
```
tgsmith61591: @charlesdrotar are we using this anymore? I can't remember. If not, we should eventually yank that out of the package itself.

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,104 @@
...
+          fi
+          pytest --showlocals --durations=20 --mpl --mpl-baseline-path=etc/pytest_images --pyargs pmdarima --benchmark-skip
```
tgsmith61591: You can change this to:

```
pytest --showlocals --durations=20 --pyargs pmdarima --benchmark-skip
```

We no longer need MPL image comparisons in our testing.

tgsmith61591: In fact, if that's still in the Azure pipelines, we should remove that too.

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,104 @@
...
+        shell: bash
+
+      # We build the source archive separately because of this: https://github.com/alkaline-ml/pmdarima/pull/136#discussion_r279781731
+      # This also implicitly runs `make version`, which -- if tagged -- gets the right version
+      - name: Building source distribution
+        run: make sdist
```
tgsmith61591: In your opinion would it be worth explicitly running `make version` before or after this just for readability? We know the `make sdist` target calls it, but every time I read through this I have a minor panic attack that we'll release 0.0.0 wheels.

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,104 @@
...
+          fi
+        shell: bash
+
+      - name: Ensuring sdist can be installed
+        run: pip install dist/$(ls dist | grep tar)
+        shell: bash
+
+      # Use this one for testing: python -m twine upload --repository-url https://test.pypi.org/legacy/ --skip-existing dist/pmdarima-*
+      - name: Deploying to PyPI
+        if: github.event_name == 'tag' && success()  # Only deploy on tags
```
tgsmith61591: Should we also add `&& [ -f pmdarima/VERSION ]` (or whatever equivalent) so we guarantee we only release if the VERSION file was created?

build_tools/azure/render_meta.py
```diff
@@ -0,0 +1,52 @@
...
+import sys
+
+from jinja2 import Environment, FileSystemLoader
+
+# Constant file paths. Relative to this file, so we have to run this file from this directory
+DIST_PATH = '../../dist'
+VERSION_FILE = '../../VERSION'
+REQUIREMENTS_FILE = '../../requirements.txt'
+OUTPUT_DIR = '../../conda'
+OUTPUT_FILE = '../../conda/meta.yaml'  # conda is weird about yml vs yaml, so we have to use yaml
```
tgsmith61591: I get a little bit of heartburn seeing rel paths only. Someone will invariably mess up someday. Could we add:

```python
ABSDIR = os.path.abspath(os.path.dirname(__file__))
```

And make them all, e.g.,

```python
DIST_PATH = ABSDIR + "../../dist"
...
```

build_tools/azure/render_meta.py
```diff
@@ -0,0 +1,52 @@
+import os
+import sys
+
+from jinja2 import Environment, FileSystemLoader
+
+# Constant file paths. Relative to this file, so we have to run this file from this directory
+DIST_PATH = '../../dist'
+VERSION_FILE = '../../VERSION'
```
tgsmith61591: Shouldn't this point into the `pmdarima` dir?

build_tools/get_tag.py
```diff
@@ -29,6 +29,13 @@ def get_version_from_tag(tag):
         tag = os.getenv('BUILD_SOURCEBRANCH').split('/')[-1]
         f.write(get_version_from_tag(tag))
 
+elif os.getenv('GITHUB_REF') and \
+        os.getenv('GITHUB_REF').startswith('refs/tags/'):
```
tgsmith61591: Does GH really use the `refs/tags` that Azure does? That's convenient.

.github/workflows/build_and_deploy.yml
aaronreidsmith: My bad, that was a comment from when I couldn't get 3.5 to work properly on GitHub. I will remove that

.github/workflows/build_and_deploy.yml
aaronreidsmith: This is a quirk of GitHub Actions. They are essentially trying to make it so you can use other people's action in your workflow. In this case, to set up the correct Python (Version and Architecture), we have to use the [setup-python action](https://github.com/actions/setup-python). This is a first-party action from GitHub, so it should be safe

build_tools/azure/render_meta.py
aaronreidsmith: Yeah, I can definitely do that. `pathlib` was giving me trouble on 3.5 on GitHub Actions, so this is an artifact of that. I can even try to update it to use `pathlib` if we want, since this is now on ADO

build_tools/get_tag.py
aaronreidsmith: [They sure do!](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables#default-environment-variables)

build_tools/azure/render_meta.py
aaronreidsmith: Yeah, this is another reason relative paths suuuuck. I moved this file and forgot to update that one. Will fix

.github/workflows/build_and_deploy.yml
aaronreidsmith: Also, 3.8 PyPI distributions will likely git built on GitHub (at least for mac and windows). I'd like to keep it as:

- CircleCI - Linux
- Github Actions - Windows and Max
- Azure Pipelines - Conda

.github/workflows/build_and_deploy.yml
aaronreidsmith: Yeah, it doesn't hurt to be explicit! I will add that

.github/workflows/build_and_deploy.yml
aaronreidsmith: Fixed on ADO (conda) and Github Actions with 006a582415861cbfe9bac8104990770b316bfdce

.github/workflows/build_and_deploy.yml
aaronreidsmith: Fixed with 575849598b04ddd728cbb702c9e845ab00e6bad6

build_tools/azure/render_meta.py
aaronreidsmith: I changed this to use `pathlib`. Fixed with 006a582415861cbfe9bac8104990770b316bfdce

build_tools/azure/render_meta.py
aaronreidsmith: This should be fixed with 575849598b04ddd728cbb702c9e845ab00e6bad6 
 and 006a582415861cbfe9bac8104990770b316bfdce. I'll leave this unresolved so you can take a look

.github/workflows/build_and_deploy.yml
aaronreidsmith: Added in b39e1b73be3b2709da4527be5eb4d59e24abbf0d. We have to do a check in a previous step, then store output. I confirmed it [here](https://github.com/alkaline-ml/pmdarima/pull/263/checks?check_run_id=345830965#step:14:1)

.github/workflows/build_and_deploy.yml
charlesdrotar: Looks like we are using it in [visualization.py](https://github.com/alkaline-ml/pmdarima/blob/master/pmdarima/utils/visualization.py#L21-L23). Strangely the comment says something about Travis builds in particular and `mpl` seems to be in all the visualization functions so we would need to investigate if we should or shouldn't do a clean yank.

charlesdrotar: That said assuming we don't use the environment variable and we aren't using the debug flag in [matplotlib.py's get_compatible_pyplot()](https://github.com/alkaline-ml/pmdarima/blob/master/pmdarima/compat/matplotlib.py#L17), we are basically just doing a normal import of matplotlib.

build_tools/azure/conda/linux.yml
```diff
@@ -0,0 +1,30 @@
+steps:
```
charlesdrotar: **Nit:** I see a lot of these steps being candidates for a template so that it can be abstracted and shared across multiple OS-specific yamls.

.github/workflows/build_and_deploy.yml
aaronreidsmith: Travis is where we used to build mac, so it probably should have said mac machine. I'll let more knowledgable people decide if we should yank it or not

build_tools/azure/conda/linux.yml
aaronreidsmith: I agree in principle, but in practice, it was a little harder. Each OS has its own quirks:

1. macOS requires `sudo` for setting up conda, but otherwise is the same as Linux
2. Linux and macOS have to build a wheel _before_ building conda
3. Windows requires a little bit of powershell, and to build conda from source.

That is the [way I initially had it](https://dev.azure.com/tgsmith61591gh/pmdarima/_build/results?buildId=1190), but I didn't really like how there were 2 or 3 OS-specific lines for some steps. Happy to change it if y'all prefer it that way!

.github/workflows/build_and_deploy.yml
tgsmith61591: Aaron is correct. This is from the days of yore when we tested viz utils by comparing to PNG files. It can probably be removed at some later point.

build_tools/azure/render_meta.py
```diff
@@ -0,0 +1,58 @@
...
+    autoescape=False,
+    loader=FileSystemLoader(str(TEMPLATE_PATH.resolve())),
+    trim_blocks=False
+)
+
+# Find the version
+try:
+    VERSION = open(str(VERSION_FILE.resolve())).readline().strip()
+except FileNotFoundError:
+    VERSION = '0.0.0'
```
tgsmith61591: Can/should we raise here? I would really hope that come Conda release time we have a `VERSION` file :)

aaronreidsmith: I agree we shouldn't release if we can't find the VERSION file, however I don't think it can happen here. Conda's `meta.yaml` file is required to build the package, and `version` is a required key in that file. So this is required to build the conda distribution even if we have no intention of distributing it. I can add something similar to what I did [here](https://github.com/alkaline-ml/pmdarima/pull/263#discussion_r357141289) on Github Actions

aaronreidsmith: This is handled as of 70cdd4f. See [here](https://dev.azure.com/tgsmith61591gh/pmdarima/_build/results?buildId=1415&view=logs&j=2e703997-b85c-5955-fe15-cdf93bef6757&t=e51ec69b-f0fa-593a-ddf4-8cce6a79d81c&l=2)

aaronreidsmith:
3 final things on this PR:

1. We need a conda channel/token to deploy
2. Do we like have 3 separate files for the conda builds or do want to combine them and have conditional logic?
3. Do we like the Github Actions badge? We can change/remove the name if desired:

Github Badge

CC: @charlesdrotar @tgsmith61591

PR #264, opened 2019-12-12
[MRG+2] Seasonal decomposition example docs

charlesdrotar:
# Description

Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context. List any dependencies
that are required for this change.

Fixes #(issue)

## Type of change

- [X] Documentation change

# How Has This Been Tested?

for your test configuration

- [X] Ran `make documentation` and spot checked locally.

# Checklist:


examples/arima/example_seasonal_decomposition.py
```diff
@@ -0,0 +1,79 @@
...
+
+   <br/>
+"""
+print(__doc__)
+
+# Author: Charles Drotar <drotarcharles@gmail.com>
+
+from pmdarima.arima.seasonality import decompose
+from pmdarima.datasets import load_ausbeer, load_airpassengers
+from pmdarima.utils.visualization import decomposed_plot
```
tgsmith61591: RST is weird. It will only find the link to the `decompose` doc at the module level. Can you make the imports:

```python
from pmdarima import arima
from pmdarima import datasets
from pmdarima import utils
```

And then refer to the methods by `<namespace>.<method>`? Like `utils.decomposed_plot`. This will create a link in the actual method documentation to any examples that use it at the module level.

charlesdrotar: Will do!

charlesdrotar: Resolved as best I could but it seems to be a bit of a buggy feature in that it doesn’t render all correctly despite making the change.

PR #269, opened 2019-12-15
[MRG+2] Host Pypy dependencies ourselves

aaronreidsmith:
# Description

This PR helps us as we move toward Python 3.8 support. It requires us to update some of our requirements, which do not have pre-built wheels for pypy. This PR allows us to build new pypy wheels whenever `requirements.txt` is updated so that they can be used for pypy builds on CircleCI.

#### TODO:
- [x] Create an `alkaline-ml` Gemfury account (@tgsmith61591)
- [x] Change the YAML file to push to this account instead of my personal account
- [x] Uncomment trigger logic so it only triggers on changes to `requirements.txt` and the YAML file itself (has to wait until they are pushed to the new Gemfury account) 
- [ ] Update [alkaline-ml/pmdarima-docker](https://github.com/alkaline-ml/pmdarima-docker/blob/master/circle-pypy-base/Dockerfile) to point to our self-hosted index

## Type of change

# How Has This Been Tested?

- [X] Wheels are downloadable and installable from Gemfury

# Checklist:


aaronreidsmith:
This one is good-to-go now. All of our dependencies have been uploaded to Gemfury. You can test it by installing one of the dependencies that is not necessarily pypy-specific:

```bash
asmith ~ $ pip install --index-url https://pypi.fury.io/alkaline-ml joblib
Looking in indexes: https://pypi.fury.io/alkaline-ml
Collecting joblib
  Downloading https://pypi.fury.io/alkaline-ml/-/ver_2hQTpn/joblib-0.14.1-py2.py3-none-any.whl (294kB)
     |████████████████████████████████| 296kB 1.1MB/s 
Installing collected packages: joblib
Successfully installed joblib-0.14.1
```

Screen Shot 2019-12-16 at 1 50 10 PM
Screen Shot 2019-12-16 at 1 50 15 PM

`pmdarima` was uploaded as a test, but can only be deleted by the owner (@tgsmith61591)

aaronreidsmith:
After the release of Scipy 1.4.0, I had to do a lot of messing around with this one. Should be good, if you want to take a look now @tgsmith61591

.github/workflows/build_pypy_dependencies.yml
```diff
@@ -0,0 +1,66 @@
+name: Build PyPy Dependencies
+
+on:
+  push:
+    paths:
+    - 'requirements.txt'
+    - '.github/workflows/build_pypy_dependencies.yml'
```
tgsmith61591: Do we really want to run this every push? Would it make more sense for us to put this into `pmdarima-docker` or somewhere else?

tgsmith61591: OH, I think I understand. This only runs on a push where those paths have been updated. That's pretty slick.

tgsmith61591: Can you also add `setup.py`?

aaronreidsmith: Yeah, you got it! I intentionally do it on just pushes so that if a contributor updates the requirements, by PR time the dependencies should be built and the Circle pypy job should be quicker. Went ahead and added `setup.py`

PR #283, opened 2019-12-26
[MRG+2] Test deployments

aaronreidsmith:
# Description

This change allows us to test our deployment logic by tagging with _almost any_ tag that (contains a letter)  Examples:

- `v1.5.3rc`
- `v1.5.3rc1`
- `v1.5.3test`

If we tag in this way, the CI/CD pipelines will deploy to:

- test.pypi.org (for PyPI releases)
- Our anaconda.org channel with a label of `test` 
  - Adding this label means no one will see it unless they are specifically searching for it ([source](https://docs.anaconda.com/anaconda-cloud/user-guide/tasks/work-with-packages/#uploading-packages))

It also checks to make sure if the version doesn't match the test format, that it matches one of the following before deploying to production:

- `X.Y.Z`
- `X.Y`

I also removed the `fail-fast = False` logic from GitHub Actions, as that is mostly for debugging.

## Type of change

# How Has This Been Tested?

- [X] Tested the bash regex logically locally and it works great

# Checklist:

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -130,11 +130,11 @@ jobs:
           fi
         shell: bash
 
-      # Use this one for testing: python -m twine upload --repository-url https://test.pypi.org/legacy/ --skip-existing dist/pmdarima-*
       - name: Deploying to PyPI
         # Only deploy on tags and when VERSION file created
         if: github.event_name == 'tag' && success() && steps.version_check.version_exists == 'true'
         run: |
+          # If the VERSION ends with a letter (beta, rc, etc.), it is considered a test release
           if [[ $(cat ${GITHUB_WORKSPACE}/pmdarima/VERSION) =~ '[a-zA-Z]'$ ]]; then
             python -m twine upload --repository-url https://test.pypi.org/legacy/ --skip-existing dist/pmdarima-*
           else
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -133,11 +133,15 @@ jobs:
         # Only deploy on tags and when VERSION file created
         if: github.event_name == 'tag' && success() && steps.version_check.version_exists == 'true'
         run: |
-          # If the VERSION ends with a letter (beta, rc, etc.), it is considered a test release
-          if [[ $(cat ${GITHUB_WORKSPACE}/pmdarima/VERSION) =~ '[a-zA-Z]'$ ]]; then
+          # Check our VERSION. Basically, if it contains letters, it is a pre-release. Othewise,
+          # it has to match X.Y or X.Y.Z
+          if [[ $(cat ${GITHUB_WORKSPACE}/pmdarima/VERSION) =~ '^[0-9]+\.[0-9]+\.?[0-9]*?[a-zA-Z]+[0-9]*$' ]]; then
             python -m twine upload --repository-url https://test.pypi.org/legacy/ --skip-existing dist/pmdarima-*
-          else
+          elif [[ $(cat ${GITHUB_WORKSPACE}/pmdarima/VERSION) =~ '^[0-9]+\.[0-9]+\.?[0-9]*?$' ]]; then
             python -m twine upload --skip-existing dist/pmdarima-*
+          else
+            echo 'Malformed tag'
+            exit 1
           fi
         shell: bash
         env:
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -134,7 +134,12 @@ jobs:
       - name: Deploying to PyPI
         # Only deploy on tags and when VERSION file created
         if: github.event_name == 'tag' && success() && steps.version_check.version_exists == 'true'
-        run: python -m twine upload --skip-existing dist/pmdarima-*
+        run: |
+          if [[ $(cat ${GITHUB_WORKSPACE}/pmdarima/VERSION) =~ '[a-zA-Z]'$ ]]; then
+            python -m twine upload --repository-url https://test.pypi.org/legacy/ --skip-existing dist/pmdarima-*
+          else
+            python -m twine upload --skip-existing dist/pmdarima-*
+          fi
         shell: bash
         env:
           TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -136,8 +136,10 @@ jobs:
           # Check our VERSION. Basically, if it contains letters, it is a pre-release. Otherwise,
           # it has to match X.Y or X.Y.Z
           if [[ $(cat ${GITHUB_WORKSPACE}/pmdarima/VERSION) =~ '^[0-9]+\.[0-9]+\.?[0-9]*?[a-zA-Z]+[0-9]*$' ]]; then
+            echo 'Uploading to test pypi'
             python -m twine upload --repository-url https://test.pypi.org/legacy/ --skip-existing dist/pmdarima-*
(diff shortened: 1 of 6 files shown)


build_tools/circle/build_push_doc.sh
```diff
@@ -13,7 +13,7 @@ fi
 
 # The version is retrieved from the CIRCLE_TAG. If there is no version, we just
 # call it 0.0.0, since we won't be pushing anyways (not master and no tag)
-if [[ ! -z ${CIRCLE_TAG} ]]; then
+if [[ ! -z ${CIRCLE_TAG} && ${CIRCLE_TAG} =~ '^v?[0-9]+\.[0-9]+\.?[0-9]*?[a-zA-Z]+[0-9]*$' ]]; then
```
aaronreidsmith: @tgsmith61591 Is this okay? This way, it _should_ skip doc deployments on non-prod tags. You know this step better than me, so I want to confirm it looks okay

aaronreidsmith: Disregard. I moved this check [here](https://github.com/alkaline-ml/pmdarima/pull/283/files#diff-edb3eabb049d60b13c2b45c3e5ece16fR163)

.github/workflows/build_and_deploy.yml
```diff
@@ -130,11 +129,20 @@ jobs:
...
 
-      # Use this one for testing: python -m twine upload --repository-url https://test.pypi.org/legacy/ --skip-existing dist/pmdarima-*
       - name: Deploying to PyPI
         # Only deploy on tags and when VERSION file created
         if: github.event_name == 'tag' && success() && steps.version_check.version_exists == 'true'
-        run: python -m twine upload --skip-existing dist/pmdarima-*
+        run: |
+          # Check our VERSION. Basically, if it contains letters, it is a pre-release. Otherwise,
+          # it has to match X.Y or X.Y.Z
+          if [[ $(cat ${GITHUB_WORKSPACE}/pmdarima/VERSION) =~ '^[0-9]+\.[0-9]+\.?[0-9]*?[a-zA-Z]+[0-9]*$' ]]; then
```
tgsmith61591: Currently only something like `1.0.3rc1` is caught by this. What if someone tags with something like `1.0.3-rc1`? Should we make more flexible like:

`^[0-9]+\.[0-9]+\.?[0-9]*?-?[a-zA-Z]+[0-9]*$`

tgsmith61591: Or does it make sense to very strictly enforce the current supported regex?

tgsmith61591: Also our `get_tag.py` supports the potential of a tag leading with `v`. Should we make the regex start with `^v?`?

.github/workflows/build_and_deploy.yml
```diff
@@ -130,11 +129,20 @@ jobs:
...
+            python -m twine upload --repository-url https://test.pypi.org/legacy/ --skip-existing dist/pmdarima-*
+          elif [[ $(cat ${GITHUB_WORKSPACE}/pmdarima/VERSION) =~ '^[0-9]+\.[0-9]+\.?[0-9]*?$' ]]; then
```
tgsmith61591: Same comment here RE: `^v?`

build_tools/azure/conda/deploy.yml
```diff
@@ -10,8 +10,21 @@ steps:
...
       # This looks like it is running `conda build` again, but it just returns the output file
       output_file=$(conda-build --output --python=$(python.version) conda/)
-      # --skip just means skip existing versions
-      anaconda upload --skip $output_file
+
+      # Check our VERSION. Basically, if it contains letters, it is a pre-release. Otherwise,
+      # it has to match X.Y or X.Y.Z
+      #
+      # --skip means skip existing (can be a little confusing)
+      if [[ $(cat ${BUILD_SOURCESDIRECTORY}/pmdarima/VERSION) =~ '^[0-9]+\.[0-9]+\.?[0-9]*?[a-zA-Z]+[0-9]*$' ]]; then
```
tgsmith61591: Check on this regex, depending on what's decided above

build_tools/azure/conda/deploy.yml
```diff
@@ -10,8 +10,21 @@ steps:
...
+        # Adding the `test` label means it doesn't show up unless you specifically
+        # search for packages with the label `test`
+        anaconda upload --label test --skip $output_file
+      elif [[ $(cat ${BUILD_SOURCESDIRECTORY}/pmdarima/VERSION) =~ '^[0-9]+\.[0-9]+\.?[0-9]*?$' ]]; then
```
tgsmith61591: and here

build_tools/circle/build_push_doc.sh
```diff
@@ -160,7 +160,7 @@ ls -la
 # Finally, deploy the branch, but if it's a pull request or tag, don't!!
 if [[ ! -z ${CIRCLE_PULL_REQUEST} ]]; then
   echo "Will not deploy doc on pull request (${CIRCLE_PULL_REQUEST})"
-elif [[ ${CIRCLE_BRANCH} == "master" || ! -z ${CIRCLE_TAG} ]]; then
+elif [[ ${CIRCLE_BRANCH} == "master" || (! -z ${CIRCLE_TAG} && ${CIRCLE_TAG} =~ '^v?[0-9]+\.[0-9]+\.?[0-9]*?[a-zA-Z]+[0-9]*$') ]]; then
```
tgsmith61591: And here

tgsmith61591: Does this need to be

```
((! -z ${CIRCLE_TAG}) && (${CIRCLE_TAG} =~ '^v?[0-9]+\.[0-9]+\.?[0-9]*?[a-zA-Z]+[0-9]*$'))
```

? Just like being liberal with parens as I am wary of order of operations in this stuff. Also, maybe update regex depending on the comments above.

build_tools/circle/deploy.sh
```diff
@@ -3,5 +3,16 @@
...
-twine upload --skip-existing dist/pmdarima-*
-# twine upload --skip-existing --repository-url https://test.pypi.org/legacy/ dist/pmdarima-*
+
+# Check our VERSION. Basically, if it contains letters, it is a pre-release. Otherwise,
+# it has to match X.Y or X.Y.Z
+#
+# On CircleCI, we look for the `v` at the beginning of the version, since we are looking at the tag
+if [[ ${CIRCLE_TAG} =~ '^v?[0-9]+\.[0-9]+\.?[0-9]*?[a-zA-Z]+[0-9]*$' ]]; then
+  twine upload --skip-existing --repository-url https://test.pypi.org/legacy/ dist/pmdarima-*
+elif [[ ${CIRCLE_TAG} =~ '^v?[0-9]+\.[0-9]+\.?[0-9]*?$' ]]; then
```
tgsmith61591: Check these ^

.github/workflows/build_and_deploy.yml
aaronreidsmith: RE: echoing, that is easy enough, I can add that.

RE: Regex, I am open to making it more flexible, so if you would like it to support hyphens, I am all for that. Technically, there is a published way these should look ([PEP 440](https://www.python.org/dev/peps/pep-0440/#public-version-identifiers)), which I have mostly covered, but I can add support for the `post` and `dev` tags

RE: Leading V, I don't think we need to add that on Azure or Github. The reason being, we are reading the actual `VERSION` file, which translates `v1.5.3` to `1.5.3`. On the Circle job, since we use the **tag** and not the `VERSION` file, you can see we _do_ support the leading V

build_tools/circle/build_push_doc.sh
aaronreidsmith: I can update that for sure. Also, I am going to do a second pass after this PR. I think technically `! -z` should be replaced with `-n` ([source](https://github.com/koalaman/shellcheck/wiki/SC2236)), and we have a lot of occurrences of that.

PR #302, opened 2020-02-24
[MRG+1] Nightly cron

aaronreidsmith:
# Description

This PR adds a nightly cron job to build `pmdarima` against all OS/Python versions we support using the **latest** version of all of our dependencies. It only builds wheel files, because if wheel files break, conda and source will break as well.

This PR adds the GitHub Actions workflow file, as well as a Python file used to find the latest versions of our dependencies.

I am planning on using [this action](https://github.com/marketplace/actions/slack-notify) to post to a slack channel if/when it fails, hence why this is a WIP (need to get all the slack secrets added)

## Type of change

# How Has This Been Tested?

- [X] The Python script successfully strips off pinned versions
- [X] The tests successfully run against latest versions

# Checklist:


aaronreidsmith:
Per our Slack conversation, this has been updated to only run against latest everything (meaning latest OS, latest Python [currently 3.8], and latest dependencies)

.github/workflows/nightly_cron.yml
```diff
@@ -0,0 +1,79 @@
...
+        os: [windows-latest, macos-latest, ubuntu-latest]
+        python-version: ['3.8']
+        architecture: ['x86', 'x64']
+        exclude:
+          # Don't build 32-bit on Mac or Linux
+          - os: macos-latest
+            architecture: x86
+
+          - os: ubuntu-latest
+            architecture: x86
```
tgsmith61591: An aside, I really like the way GH actions structures their YAML. Very readable in plain English.

.github/workflows/nightly_cron.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Installing generated wheel
+        run: pip install --pre --no-index --find-links dist/ pmdarima
+        shell: bash
+
+      - name: Running unit tests
+        run: |
+          if [ "${{ matrix.os }}" == "macos-latest" ]; then
+            export PMD_MPL_BACKEND=TkAgg
+          fi
+          pytest --showlocals --durations=20 --pyargs pmdarima --benchmark-skip
```
tgsmith61591: FWIW, you prob don't need the `--benchmark-skip` since we no longer use pytest-benchmark.

build_tools/github/get_latest_dependencies.py
```diff
@@ -0,0 +1,42 @@
...
+    requirements = []
+    for requirements_file in requirements_files:
+        with open(requirements_file) as file:
+            for line in file:
+                requirement = line.strip()
+                if line.startswith('#'):
+                    continue
+                match = re.match(r'^([A-Za-z\-0-9]+)', requirement)
+                if match.group(0).lower() not in requirements:
+                    requirements.append(match.group(0).lower())
```
tgsmith61591: I'm not super clear on this regex. But I get the purpose of the loop. FWIW (and it makes no real difference here because the size is so small and is not user-facing), set lookups are O(1) while list lookups are O(N)

aaronreidsmith: I initially had a set, but our requirements file has a specific order that I want to maintain. It's important to install Cython and Numpy before anything else, and if I use a set and cast to a list, I can't guarantee that 😕

.github/workflows/nightly_cron.yml
aaronreidsmith: Cool, I will remove that!

.github/workflows/nightly_cron.yml
aaronreidsmith: Yeah, they make it pretty simple! Wish their docs were a little better though 🤔

build_tools/github/get_latest_dependencies.py
aaronreidsmith: Also, what this regex is doing is looking for any letter characters, dashes, or numbers starting at the beginning of the string. So, it will either match anything up until our pins (`>,<,~,=`), or until the end of the line

aaronreidsmith:
Personally, I wouldn't put a badge on the readme, because I feel like we wouldn't be able to explain what it is actually representing in such a small space (the badge title). Also, since the distributions _will_ have pinned versions, it is kind of a moot point to display that. I think once I add the slack notifications, it should be good enough

aaronreidsmith:
This should be good, now that I have added the notification step

Issue #308, opened 2020-03-09
Deprecated matplotlib.pyplot.hist keyword argument "normed=true" used in pmdarima.arima.ARIMA.plot_diagnostics


adubpak:
**Describe the bug**
Using the pmdarima.arima.ARIMA.plot_diagnostics always seems to return the following error:  "AttributeError: 'Rectangle' object has no property 'normed'"

**To Reproduce**  
Use the plot_diagnostics method

```python
sarima_model = auto_arima(
    train,
    m=13,
    suppress_warnings=True,
    error_action='warn'
)

sarima_model.plot_diagnostics(lags=13)
```

**Versions**
System:
     python: 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)]
     machine: Windows-10-10.0.17763-SP0

Python dependencies:
     pip: 20.0.2
     setuptools: 46.0.0
     sklearn: 0.22.2.post1
     statsmodels: 0.11.1
     numpy: 1.18.1
     scipy: 1.4.1
     Cython: 0.29.15
     pandas: 1.0.1
     joblib: 0.14.1
     pmdarima: 1.5.3

**Expected behavior**
Return the plot_diagnostics figure (residual, histogram, normal Q-Q and Correlogram)

**Actual behavior**
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-44-5a958117d242> in <module>
----> 1 sarima_model.plot_diagnostics(lags=13)

c:\users\XXXXXXXX\XXXXXXXX\XXXXXXXX\venv\lib\site-packages\pmdarima\utils\metaestimators.py in <lambda>(*args, **kwargs)
     51 
     52         # lambda, but not partial, allows help() to work with update_wrapper
---> 53         out = (lambda *args, **kwargs: self.fn(obj, *args, **kwargs))
     54         # update the docstring of the returned function
     55         update_wrapper(out, self.fn)

c:\users\XXXXXXXX\XXXXXXXX\XXXXXXXX\venv\lib\site-packages\pmdarima\arima\arima.py in plot_diagnostics(self, variable, lags, fig, figsize)
   1198         # hist needs to use `density` in future when minimum matplotlib has it
   1199         with warnings.catch_warnings(record=True):
-> 1200             ax.hist(resid_nonmissing, normed=True, label='Hist')
   1201 
[...]

PR #309, opened 2020-03-10
[MRG+2] Replaced keyword argument "normed" to "density" in method pmdarima.arima.ARIMA.plot_diagnostics

adubpak:
In response to issue #308.
The "normed" keyword used in method pmdarima.arima.ARIMA.plot_diagnostics (in reference to matplotlib.pyplot.hist) is deprecated.
It has been removed from matplolib since version 3.2.0, and prevented the plot_diagnostics method from working.
It has been replaced with the new supported keyword argument "density".

### DIFF
--- a/pmdarima/arima/arima.py
+++ b/pmdarima/arima/arima.py
@@ -21,6 +21,7 @@ from ..base import BaseARIMA
 from ..compat.numpy import DTYPE  # DTYPE for arrays
 from ..compat.sklearn import get_compatible_check_is_fitted, safe_indexing
 from ..compat import statsmodels as sm_compat
+from ..compat import matplotlib as mpl_compat
 from ..compat.matplotlib import mpl_hist_arg
 from ..utils import get_callable, if_has_delegate, is_iterable, check_endog, \
     check_exog
@@ -1197,12 +1198,10 @@ class ARIMA(BaseARIMA):
         ax = fig.add_subplot(222)
         # temporarily disable Deprecation warning, normed -> density
         # hist needs to use `density` in future when minimum matplotlib has it
-        # normed keyword arugment is no longer supported in matplotlib since version 3.2.0
+        # 'normed' argument is no longer supported in matplotlib since
+        # version 3.2.0. New function added for backwards compatibility
         with warnings.catch_warnings(record=True):
-            if mpl_hist_arg:
-                ax.hist(resid_nonmissing, density=True, label='Hist')
-            else:
-                ax.hist(resid_nonmissing, normed=True, label='Hist')
+            ax.hist(resid_nonmissing, label='Hist', **mpl_compat.mpl_hist_arg())
 
         kde = gaussian_kde(resid_nonmissing)
         xlim = (-1.96 * 2, 1.96 * 2)
--- a/pmdarima/arima/arima.py
+++ b/pmdarima/arima/arima.py
@@ -1201,7 +1201,11 @@ class ARIMA(BaseARIMA):
         # 'normed' argument is no longer supported in matplotlib since
         # version 3.2.0. New function added for backwards compatibility
         with warnings.catch_warnings(record=True):
-            ax.hist(resid_nonmissing, label='Hist', **mpl_compat.mpl_hist_arg())
+            ax.hist(
+                resid_nonmissing,
+                label='Hist',
+                **mpl_compat.mpl_hist_arg()
+            )
 
         kde = gaussian_kde(resid_nonmissing)
         xlim = (-1.96 * 2, 1.96 * 2)
--- a/pmdarima/arima/arima.py
+++ b/pmdarima/arima/arima.py
@@ -1196,8 +1196,9 @@ class ARIMA(BaseARIMA):
         ax = fig.add_subplot(222)
         # temporarily disable Deprecation warning, normed -> density
         # hist needs to use `density` in future when minimum matplotlib has it
+        # normed keyword arugment is no longer supported in matplotlib since version 3.2.0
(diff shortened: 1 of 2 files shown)


pmdarima/arima/arima.py
```diff
@@ -1196,8 +1196,9 @@ def plot_diagnostics(self, variable=0, lags=10, fig=None, figsize=None):
         ax = fig.add_subplot(222)
         # temporarily disable Deprecation warning, normed -> density
         # hist needs to use `density` in future when minimum matplotlib has it
+        # normed keyword arugment is no longer supported in matplotlib since version 3.2.0
         with warnings.catch_warnings(record=True):
-            ax.hist(resid_nonmissing, normed=True, label='Hist')
+            ax.hist(resid_nonmissing, density=True, label='Hist')
```
tgsmith61591: This will break backwards compatibility with older versions. Can we we add some sort of `ax_hist` compatibility function in [`pmdarima.compat.matplotlib`](https://github.com/alkaline-ml/pmdarima/blob/master/pmdarima/compat/matplotlib.py) that uses `density` in versions > 3.2.0 and `normed` in previous versions?

adubpak: The `density` keyword is present in matplotlib since version 2.1.0 (released on October 7 2017): https://matplotlib.org/2.1.0/api/_as_gen/matplotlib.pyplot.hist.html
It should be compatible starting form that version.

I looked at the source code of statsmodels that is called by `plot_diagnostics` to verify whether there was a requirement of the matplotlib version, but there seems to be none: it simply tries to import matplotlib, otherwise it raises an error.

Thus, you are right that this would break compatibility with previous matplotlib versions (or that a requirement would need to be added in pmdarima for `matplotlib>=2.1.0`, which is not optimal).

To resolve this issue, I took your idea and created the function `mpl_hist_arg` in `pmdarima.compat.matplotlib`, which returns `True` if `matplotlib>=2.1.0` and otherwise returns `False`.
It also returns `True` if the matplotlib version cannot be evaluated.
If `mpl_hist_arg=True`, `ax.hist` uses `density=True`, otherwise `normed=True` is used.

Since I'm relatively new to Python, there might be a more elegant/Pythonic way to acheive this, instead of repeating twice the `ax.hist` call in the code?

aaronreidsmith: Your new solution is certainly functional, but since you say you're new to Python, I hope you'll accept some constructive feedback towards a more "Pythonic" suggestion.

`normed` and `density` are what are known as [`kwargs`](https://stackoverflow.com/a/1769475/10696164) (Key-Word arguments), meaning they don't need to be supplied, but we _can_ supply them as a key-value pair. The trick is, we need to dynamically either supply `normed=True` or `density=True`. The way to do this is to supply a dictionary with the key-value pair we want, and ["unpack" it](https://www.python.org/dev/peps/pep-0448/) using `**`.

So, with that being said, I think we could re-write your compat function something like this:

```python
def mpl_hist_arg(value=True):
    """Find the appropriate `density` kwarg for our given matplotlib version.

    This will determine if we should use `normed` or `density`. Additionally,
    since this is a kwarg, the user can supply a value (True or False) that
    they would like in the output dictionary.

    Parameters
    ----------
    value : bool, optional (default=True)
        The boolean value of density/normed

    Returns
    -------
    density_kwarg : dict
        A dictionary containing the appropriate density kwarg for the
        installed  matplotlib version, mapped to the provided or default
        value
    """
    import matplotlib
    
    density_kwarg = 'density' if matplotlib.__version__ >= '2.1.0' else 'normed'
    return {density_kwarg: value}
```

Then, we can cleanly implement it like this:

```python
ax.hist(resid_nonmissing, label='Hist', **mpl_compat.mpl_hist_arg())
```

Since they are key word arguments, it doesn't matter that we changed the order. But key word arguments should generally be last as to not interfere with positional arguments. Additionally, we already have a check in `plot_diagnostics` to make sure `matplotlib` is installed, so that is why I left it out of my re-written function.

Let me know if that makes sense, but I think that would be a clean way to implement this without having to duplicate code

pmdarima/arima/arima.py
```diff
@@ -1196,8 +1197,12 @@ def plot_diagnostics(self, variable=0, lags=10, fig=None, figsize=None):
         ax = fig.add_subplot(222)
         # temporarily disable Deprecation warning, normed -> density
         # hist needs to use `density` in future when minimum matplotlib has it
+        # normed keyword arugment is no longer supported in matplotlib since version 3.2.0
```
aaronreidsmith: In addition to the discussion that is happening on the actual implementation, our linter is complaining that this line is too long (>79 characters), hence why the tests are failing

pmdarima/arima/arima.py
adubpak: Thank you for the feedback, it is quite appreciated!
I am indeed quite new to Python, I just started learning last month.
I had read about `kwargs` before, but your explanation really made me understand the concept.

I will edit the code as per your example.

Should we keep catching the deprecation warning for backwards compatibility (in case the `normed=True` is used)?
`with warnings.catch_warnings(record=True):`

pmdarima/arima/arima.py
adubpak: I will edit the line to be less <=79 characters

pmdarima/arima/arima.py
aaronreidsmith: Awesome, looks good now!

Yeah, I think catching the warnings is still good. For anyone that has an older version of `matplotlib` installed, we don't want to suddenly start raising warnings in their workflows.

PR #324, opened 2020-04-22
MAINT: Rewrite check_is_fitted

tgsmith61591:
The scikit `check_is_fitted` makes some sweeping assumptions that are a bit dangerous for us to use (anything attribute that ends with `_` and doesn't start with `__` indicates the model has been fit), so we should just use a simplified version of the old-style `check_is_fitted` that looks for specific attributes.


pmdarima/compat/sklearn.py
```diff
@@ -4,50 +4,40 @@
...
-
-    # Otherwise we use the attribute that was passed in regardless of version.
-    return check_is_fitted(estimator=estimator, attributes=attributes)
+    if isinstance(attributes, str):
+        attributes = [attributes]
+    if not hasattr(attributes, "__iter__"):
+        raise TypeError("attributes must be a string or iterable")
+    for attr in attributes:
+        if hasattr(estimator, attr):
+            return
```
aaronreidsmith, line 39: So just so I understand this, assuming this is passed an iterable of attributes, this returns after it encounters the first valid attribute? For the most part, it looks like we call this with strings, do we have an example with iterables?

tgsmith61591: That's correct. I don't have an example, but I can see a situation (particularly in a transformer) where depending on the constructor hparams, we may set either one fit attribute or another, in which case we'd want to check that either exists.

aaronreidsmith: Cool, makes sense 👍

Issue #351, opened 2020-06-26
ValueError: Found array with 0 sample(s) (shape=(0,)) while a minimum of 1 is required ?


HelgeCPH:
**Question**

I have a timeline with number of some reports per week. When I try to fit a model with `auto_arima`, I receive an error `ValueError: Found array with 0 sample(s) (shape=(0,)) while a minimum of 1 is required.` (Find the complete error trace below). 

Since I am not a statistician, I do not understand what that error tells me in relation to my data. I am supposing it has something to do with seasonality given the error trace. Since the data is per week, I set `m=52`. From my programming understanding, this seems to be the reason for the error. When choosing a smaller `m` the error disappears. But I cannot really choose another `m` due to weekly data. 

In essence my question is what does `auto_arima` try to tell me here?

Even though I can see that the error originates from the `check_array` function in sklearn's `validation.py`, perhaps that exception could be caught and instead a "domain-specific" exception could be raised by `pmdarima`?

I built a small example that reproduces my issue:

```python
import pmdarima as pm
import matplotlib.pyplot as plt


week_number = ["201751", "201752", "201801", "201802", "201803", "201804", "201805", 
               "201806", "201807", "201808", "201809", "201810", "201811", "201812", 
               "201813", "201814", "201815", "201816", "201817", "201818", "201819", 
               "201820", "201821", "201822", "201823", "201824", "201825", "201826", 
               "201827", "201828", "201829", "201830", "201831", "201832", "201833", 
               "201834", "201835", "201836", "201837", "201838", "201839", "201840", 
               "201841", "201842", "201843", "201844", "201845", "201846", "201847", 
               "201848", "201849", "201850", "201851", "201852", "201853", "201900", 
               "201901", "201902", "201903", "201904", "201905", "201906", "201907", 
               "201908", "201909", "201910", "201911", "201912", "201913", "201914", 
               "201915", "201916", "201917", "201918", "201919", "201920", "201921", 
               "201922", "201923", "201924", "201925", "201926", "201927", "201928", 
               "201929", "201930", "201931", "201932"]
reports_per_week = [1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
                    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 2, 1, 6, 2, 1, 0,
[...]

Issue #341, opened 2020-06-05
'Last 2 dimensions of the array must be square'


caxavier:
Hi,

I have been using your program pmdarima to calculate forecast models.
Here is the arguments that I use:

```python
smodel = pm.auto_arima(Train['Count'], start_p=1, start_q=1,
                               test='adf',
                               max_p=3, max_q=3, m=52,
                               start_P=0, seasonal=True,
                               d=None, D=1, trace=True,
                               error_action='ignore',  
                               suppress_warnings=True, 
                               stepwise=True)
```

Recently I've came up with an error when I run it with a specific set of data.
              
```
  File "C:\Users\carolina.xavier\AppData\Roaming\Python\Python37\site-packages\pmdarima\arima\auto.py", line 436, in auto_arima
    d = ndiffs(dx, test=test, alpha=alpha, max_d=max_d, **offset_test_args)
  File "C:\Users\carolina.xavier\AppData\Roaming\Python\Python37\site-packages\pmdarima\arima\utils.py", line 165, in ndiffs
    pval, dodiff = testfunc(x)
  File "C:\Users\carolina.xavier\AppData\Roaming\Python\Python37\site-packages\pmdarima\arima\stationarity.py", line 338, in should_diff
    res = self._ols(x, y, z, k)
  File "C:\Users\carolina.xavier\AppData\Roaming\Python\Python37\site-packages\pmdarima\arima\stationarity.py", line 296, in _ols
    return sm.OLS(yt, X, hasconst=True).fit(method='qr')
  File "C:\Users\carolina.xavier\AppData\Roaming\Python\Python37\site-packages\statsmodels\regression\linear_model.py", line 332, in fit
    beta = np.linalg.solve(R, effects)
  File "<__array_function__ internals>", line 6, in solve
  File "C:\Users\carolina.xavier\AppData\Local\Continuum\anaconda3\lib\site-packages\numpy\linalg\linalg.py", line 387, in solve
    _assert_stacked_square(a)
  File "C:\Users\carolina.xavier\AppData\Local\Continuum\anaconda3\lib\site-packages\numpy\linalg\linalg.py", **line 214, in _assert_stacked_square
    raise LinAlgError('Last 2 dimensions of the array must be square')
numpy.linalg.LinAlgError: Last 2 dimensions of the array must be square**
```

The data is weekly an has 1 year and 4 months of data represented (1-Jan-2015 to 30-April-2020).
The train has length = 56.

[...]

PR #352, opened 2020-06-29
[MRG+1] Emit warnings / custom errors for differencing issues

tgsmith61591:
# Description

Two recent issues pointed out areas where a large enough `m` could seasonally difference an array into being too small for fitting downstream models. This PR adds a custom error for #341 and a warning for #351 

## Type of change

- [X] Documentation change

# How Has This Been Tested?

Added new tests that capture the warning / custom error and assert on them.

# Checklist:

### DIFF
--- a/pmdarima/arima/tests/test_auto.py
+++ b/pmdarima/arima/tests/test_auto.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from pmdarima.arima import auto
+from pmdarima.compat.pytest import pytest_error_str
 import pytest
 
 
@@ -25,7 +26,7 @@ def test_issue_341():
          191, 234, 253, 155, 25, 0, 228, 234, 265, 205, 191, 19, 0, 188,
          156, 172, 173, 166, 28, 0, 209, 160, 159, 129, 124, 18, 0, 155]
 
-    try:
+    with pytest.raises(ValueError) as ve:
         auto.auto_arima(
             y,
             start_p=1,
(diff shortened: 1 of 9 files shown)


pmdarima/arima/tests/test_auto.py
```diff
@@ -17,3 +17,36 @@ def test_deprecation_warnings_on_class():
...
+            D=1,
+            trace=True,
+            error_action='ignore',
+            suppress_warnings=True,
+            stepwise=True
+        )
+
+    # assert that we catch the np LinAlg error and reraise with a more
+    # meaningful message
+    except ValueError as v:
```
aaronreidsmith, line 49: I think the [preferred way](https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions) to do this in pytest is like this:

```python
with pytest.raises(ValueError) as error:
    auto.auto_arima(
        y,
        start_p=1,
        start_q=1,
        test='adf',
        max_p=3,
        max_q=3,
        m=52,
        start_P=0,
        seasonal=True,
        d=None,
        D=1,
        trace=True,
        error_action='ignore',
        suppress_warnings=True,
        stepwise=True
    )

assert "Encountered exception in stationarity test" in str(error.value)
```

aaronreidsmith: Or I guess you could use `pmdarima.compat.pytest.pytest_error_str` as we do in `test_utils.py`

pmdarima/tests/test_context_managers.py
```diff
@@ -0,0 +1,52 @@
...
+    try:
+        with ctx.except_and_reraise(
+                ValueError,
+                raise_err=KeyError,
+                raise_msg="bar message"
+        ):
+            raise ValueError("contains foo message")
+    except KeyError as k:
+        assert "bar message" in str(k)
+        assert "raised from ValueError" in str(k)
```
aaronreidsmith, line 19: Same as my other comment

pmdarima/arima/tests/test_auto.py
tgsmith61591: Yes, we should. Not sure why I didn't do that here.

PR #356, opened 2020-07-07
[MRG] Address issue #354

### LINKED ISSUE #354: Checking for near non-invertible roots in auto_arima is subsequently ignored
**Describe the bug**
Checking for near non-invertible roots occurs in `_auto_solvers._root_test(model, ic, trace)` - this returns `np.inf` if the roots are near non-invertible. 

However, the return value from `_root_test` is only used to determine which model should be saved as `.bestfit` during a stepwise parameter search. The actual model IC remains unmodified and the model itself is stored in `.results_dict` for onward calculation. In the non-stepwise approach, this test is not performed at all.

Once the parameter search has finished, all the results are fitltered for None in `auto._post_ppc_arima` and then sorted according to the IC. Assuming only one model is required, `auto._return_wrapper` returns the value at the top of the list. The root testing is therefore ignored as the only possible effect it can have is on the model stored in `.bestfit`, which is not used outside the stepwise calculation.

In the R version of the code, it appears that the root testing modifies the IC of the actual model (setting it to Inf) and therefore the model will be ignored in subsequent calculations.

**To Reproduce**  
Run a stepwise parameter search, where the model with the lowest AIC has near non-invertible roots

**Versions**
Darwin-18.7.0-x86_64-i386-64bit
Python 3.7.4 (default, Oct 12 2019, 18:55:28) 
[Clang 11.0.0 (clang-1100.0.33.8)]
pmdarima 1.5.3
NumPy 1.18.2
SciPy 1.4.1
Scikit-Learn 0.22.2.post1
Statsmodels 0.11.1

**Expected behavior**
All models with near non-invertible roots should be discarded. Of the remaining models, the model with the lowest AIC should be selected

**Actual behavior**
The model with the lowest AIC is selected, regardless of whether the roots are near non-invertible

**Additional context**
Having a quick look at the R code, it seems that the root testing modifies the IC of the model itself, which would avoid this issue occurring ( https://github.com/robjhyndman/forecast/blob/master/R/newarima2.R#L780 )

Incidentally, the final model order on the stepwise search is reported from the `.bestfit` paramater of the model ( https://github.com/alkaline-ml/pmdarima/blob/master/pmdarima/arima/_auto_solvers.py#L342 ) but this may not reflect the final values selected.

tgsmith61591:
# Description

Issue #354 pointed out a few corner cases where models with near non-invertable roots could still be considered as candidate models. This PR addresses those cases by filtering them from the returned stepwise search candidates.

Fixes #354 

## Type of change

- [X] Documentation change

# How Has This Been Tested?

- [X] Test the new `_sort_and_filter` method
- [X] All existing tests still pass
- [ ] (TODO) Mock corner cases pointed out in the issue and assert they no longer return bad model candidates

# Checklist:

### DIFF
--- a/doc/no-successful-model.rst
+++ b/doc/no-successful-model.rst
@@ -10,7 +10,7 @@ For certain time series, the search may return no viable models::
       File "<stdin>", line 1, in <module>
         "Could not successfully fit a viable ARIMA model "
     ValueError: Could not successfully fit a viable ARIMA model to input data using the stepwise algorithm.
-    See http://alkaline-ml.com/pmdarima/no-successful-model.htmlfor more information on why this can happen.
+    See http://alkaline-ml.com/pmdarima/no-successful-model.html for more information on why this can happen.
 
 
 This can happen for a number of reasons:
--- a/doc/no-successful-model.rst
+++ b/doc/no-successful-model.rst
@@ -1,0 +1,29 @@
+.. _no_successful_model:
+
+===================================
+When no viable models can be found
+===================================
+
+For certain time series, the search may return no viable models::
+
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
+        "Could not successfully fit a viable ARIMA model "
+    ValueError: Could not successfully fit a viable ARIMA model to input data using the stepwise algorithm.
+    See http://alkaline-ml.com/pmdarima/no-successful-model.htmlfor more information on why this can happen.
+
+
+This can happen for a number of reasons:
+
+* Most commonly, the roots of your model may be nearly non-invertible, meaning the inverted roots
+  lie too close to the unit circle. Here's a good `blog post <https://robjhyndman.com/hyndsight/arma-roots/>`_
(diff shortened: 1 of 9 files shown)


doc/no-successful-model.rst
```diff
@@ -0,0 +1,29 @@
...
+  on the subject. Make sure ``trace`` is truthy in order to see these warnings when fitting your model.
+
+* Sometimes, your data may not be stationary and can raise errors from statsmodels when fitting. In this case,
+  the stepwise algorithm will filter out problem model fits. This can arise in a number of situations, ranging
+  from non-stationarity to actual code errors. Setting ``error_action='trace'`` will log the stacktraces of
+  any errors encountered during the search.
+
+Make sure to set ``trace`` to at least 1 in order to see the search progress, and to a value >1 to see the
+maximum trace logging available. If you still cannot diagnose why you are getting this error message, consider
+:ref:`filing_bugs`.
```
aaronreidsmith, line 29: Should this be `filing a bug` with the change above?

PR #360, opened 2020-07-10
[MRG+2] support inverse_transformer when return_conf_int=True

skyetim:
# Description

Currently, if there's an endogenous transformer in the pipeline and try to run `pipeline.predict(*, return_conf_int=True)`, pipeline would not do inverse transform. This pull request tries to fix this. 

## Type of change

# How Has This Been Tested?
Current tests cover this functionality. 

# Checklist:

### DIFF
--- a/pmdarima/pipeline.py
+++ b/pmdarima/pipeline.py
@@ -450,17 +450,18 @@ class Pipeline(BaseEstimator):
         conf_ints = None
         if return_conf_int:
             y_pred, conf_ints = y_pred
-            Xt_lower_bound, Xt_upper_bound = Xt, Xt
 
         # step through transformers in the reverse order
         for name, transformer in self.steps_[::-1]:
             if isinstance(transformer, BaseEndogTransformer):
                 y_pred, Xt = transformer.inverse_transform(y_pred, Xt)
                 if return_conf_int:
-                    conf_ints[:, 0], Xt_lower_bound = transformer.inverse_transform(
-                        conf_ints[:, 0], Xt_lower_bound)
-                    conf_ints[:, 1], Xt_upper_bound = transformer.inverse_transform(
-                        conf_ints[:, 1], Xt_upper_bound)
+                    # inverse transform of Xt is irrelevant to y
+                    # so only transform it once
+                    conf_ints[:, 0], _ = transformer.inverse_transform(
+                        conf_ints[:, 0], Xt)
+                    conf_ints[:, 1], _ = transformer.inverse_transform(
+                        conf_ints[:, 1], Xt)
 
         if return_conf_int:
             return y_pred, conf_ints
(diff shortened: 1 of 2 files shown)


pmdarima/pipeline.py
```diff
@@ -442,28 +443,24 @@ def _post_predict(self, Xt, return_vals, return_conf_int,
...
 
         # step through transformers in the reverse order
         for name, transformer in self.steps_[::-1]:
             if isinstance(transformer, BaseEndogTransformer):
                 y_pred, Xt = transformer.inverse_transform(y_pred, Xt)
+                if return_conf_int:
+                    conf_ints[:, 0], Xt_lower_bound = transformer.inverse_transform(
+                        conf_ints[:, 0], Xt_lower_bound)
+                    conf_ints[:, 1], Xt_upper_bound = transformer.inverse_transform(
+                        conf_ints[:, 1], Xt_upper_bound)
```
tgsmith61591, line 463: This is a bit tricky, since the `Xt` result of the inverse transform will either be `Xt` (returned unchanged to match the signature in the case of an endog transformer) or `Xt_prime`, which is inverse-transformed irrespective of `y`.

In either case, you shouldn't need to worry about `Xt_lower_bound` and `Xt_upper_bound` here. I think you could achieve the same thing like:
```python
                    conf_ints[:, 0], Xt = transformer.inverse_transform(
                        conf_ints[:, 0], Xt)
                    conf_ints[:, 1], _ = transformer.inverse_transform(
                        conf_ints[:, 1], Xt)
```

pmdarima/tests/test_pipeline.py
```diff
@@ -203,6 +203,9 @@ def test_pipeline_predict_inverse_transform(pipeline, exog, inverse_transform,
 
     if return_conf_ints:
```
skyetim, line 204: I believe this test already covers what I have changed. I added more detailed tests on the shape and range of conf_ints. Let me know if you think any additional tests need to be added.

pmdarima/tests/test_pipeline.py
```diff
@@ -203,6 +203,9 @@ def test_pipeline_predict_inverse_transform(pipeline, exog, inverse_transform,
...
         assert isinstance(predictions, tuple) and len(predictions) == 2
+        y_pred, conf_ints = predictions
+        assert conf_ints.shape[1] == 2
+        assert np.all((conf_ints[:, 0] <= y_pred) & (y_pred <= conf_ints[:, 1]))
```
tgsmith61591, line 208: 👍

pmdarima/pipeline.py
aaronreidsmith: @tgsmith61591 is this a typo in your response? The current implementation, does not reassign `Xt` on the first transformation, but yours does. I don't understand this enough to know which is right 🤷‍♂️

skyetim: Xt is reassigned when computing y_pred (at 458). That's the first transformation in the loop.

PR #369, opened 2020-07-31
[WIP] Add an option to return raw predictions for different horizons when using cross_val_predict

### LINKED ISSUE #364: Cross_val_predict: enable returning predictions for different horizons.
# Cross_val_predict: enable returning predictions for different horizons

I wonder if we should modify `cross_val_predict` so that it return a matrix of predictions indexed by DatetimeIndex and containing horizonwise results in columns.

## Problem
Currently `cross_val_predict` averages predictions of overlapping folds. For example, consider the case provided in [examples](http://alkaline-ml.com/pmdarima/modules/generated/pmdarima.model_selection.cross_val_predict.html#pmdarima.model_selection.cross_val_predict) with minor modifications:

```python
import pmdarima as pm
from pmdarima.model_selection import cross_val_predict, RollingForecastCV

# http://alkaline-ml.com/pmdarima/modules/generated/pmdarima.model_selection.cross_val_predict.html#pmdarima.model_selection.cross_val_predict
y = pm.datasets.load_wineind()
cv = RollingForecastCV(h=14, step=12)
preds = cross_val_predict(pm.ARIMA((1, 1, 2), 
                                   seasonal_order=(0, 1, 1, 12), 
                                   suppress_warnings=True), 
                          y, 
                          cv=cv)
```

```python
import pandas as pd
y = pd.Series(y)
y_pred = pd.Series(preds, index=y.index[-preds.shape[0]:])
y.plot();
y_pred.plot();
```

output_3_0

Currently the API returns a vector no matter how greatly folds overlap. In the above case this does not matter greatly since the overlap is relatively rare. 

However, if we were to perform an exhaustive iteration e.g. with horizon of 30 days (h=30) and step of 1  day (step=1) we would end up with significant overlap. Averaging over these results is often not the desired approach. Instead we often want to evaluate the predictive accuracy on different horizons. Like so:

| horizon (days)| accuracy |
| -------       | -------- |
| 1             |  5       |
| 2             |  6       |
| 3             |  7       |
| n             |  x       |

## Proposed solution
Getting these kind of metrics would be simple if `cross_val_predict` returned something like:

| DateTime    | t+1      | t+2     | t+3    | t+h   |
| --------    | ------ | ----- | ---- | --- |
| 01-01-1917  | 20     | 25    | 18   | p1  |
| 02-01-1917  | 13     | 17    | 29   | p2  |
| 03-01-1917  | 19     | 14    | 18   | p3  |
| ...          |   |   |   |   |

,where `t` is the forecast origin, `h` is horizon and `t+h` is naturally prediction for `h` days from `t`. E.g. in the above case we predict on January 1st of 1917 that the target variable will be 25 on January 3rd.

Looking at the source code it seems to me this wouldn't be a huge modification since a similiar matrix is calculated prior averaging. 

## Conclusion

I believe this would be of significant value because step=1 is a specially important case of crossvalidation. I am happy to help in any way possible in case @tgsmith61591 and @aaronreidsmith  feel this is a feature worth implementing. Again, thank you in advance for your input and keep up the excellent work!

tuomijal:
# Description

Currently cross_val_predict averages predictions of overlapping folds. Averaging over these results is often not the desired approach. Instead we often want to evaluate the predictive accuracy on different horizons. Following matrix as an output would enable that:

| DateTime    | t+1      | t+2     | t+3    | t+h   |
| --------    | ------ | ----- | ---- | --- |
| 01-01-1917  | 20     | 25    | 18   | p1  |
| 02-01-1917  | 13     | 17    | 29   | p2  |
| 03-01-1917  | 19     | 14    | 18   | p3  |
| ...          |   |   |   |   |

Current solution feels quite hacky, please let me know if you have ideas for a more elegant approach.

Fixes #364 

## Type of change

- [x] Documentation change

# How Has This Been Tested?

Currently only tested manually on JupyterLab.

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/pmdarima/model_selection/_validation.py
+++ b/pmdarima/model_selection/_validation.py
@@ -220,6 +220,7 @@ def cross_validate(estimator,
     return ret
 
 
+<<<<<<< HEAD
 def cross_val_predict(estimator,
                       y,
                       X=None,
@@ -227,6 +228,10 @@ def cross_val_predict(estimator,
                       verbose=0,
                       averaging="mean",
                       **kwargs):  # TODO: remove kwargs
+=======
+def cross_val_predict(estimator, y, exogenous=None, cv=None, verbose=0,
+                      averaging="mean", return_forecast_horizons=False):
+>>>>>>> first iteration
     """Generate cross-validated estimates for each input data point
 
     Parameters
@@ -266,6 +271,24 @@ def cross_val_predict(estimator,
 
         We then average each time step's forecasts to end up with our final
         prediction results.
+
+    return_forecast_horizons: bool (default=False)
+        If True, raw predictions are returned in a y x h matrix. E.g. if h=3,
+        and step=1 then:
+
+            nan nan nan # training samples
+            nan nan nan
+            nan nan nan
+            nan nan nan
+            1   4   2   # test samples
+            2   5   7
+            8   9   1
+            nan nan nan
+            nan nan nan
+
+        First column contains all one-step-ahead-predictions, second column all
+        two-step-ahead-predictions and so forth. Further metrics can then be
+        calculated as desired.
 
     Examples
     --------
@@ -326,6 +349,14 @@ def cross_val_predict(estimator,
     pred_matrix = np.ones((y.shape[0], len(prediction_blocks))) * np.nan
     for i, (pred_block, test_indices) in enumerate(prediction_blocks):
         pred_matrix[test_indices, i] = pred_block
+
+    if return_forecast_horizons:
+        pred_matrix = np.ones((y.shape[0], cv.horizon)) * np.nan
+        prediction_blocks = np.array(prediction_blocks)
+        predictions = prediction_blocks[:,0,:]
+        indices = prediction_blocks[:,1,:][:,0].astype(int) - 1
+        pred_matrix[indices] = predictions
+        return pred_matrix
 
     # from there, we need to apply nanmean (or some other metric) along rows
     # to agree on a forecast for a sample.


pmdarima/model_selection/_validation.py
```diff
@@ -250,6 +250,24 @@ def cross_val_predict(estimator, y, exogenous=None, cv=None, verbose=0,
...
+            nan nan nan
+            1   4   2   # test samples
+            2   5   7
+            8   9   1
+            nan nan nan
+            nan nan nan
+
+        First column contains all one-step-ahead-predictions, second column all
+        two-step-ahead-predictions and so forth. Further metrics can then be
+        calculated as desired.
```
tgsmith61591, line 269: 👍  Thank you for the great docstring!

pmdarima/model_selection/_validation.py
```diff
@@ -304,6 +322,14 @@ def cross_val_predict(estimator, y, exogenous=None, cv=None, verbose=0,
     for i, (pred_block, test_indices) in enumerate(prediction_blocks):
         pred_matrix[test_indices, i] = pred_block
 
+    if return_forecast_horizons:
+        pred_matrix = np.ones((y.shape[0], cv.horizon)) * np.nan
+        prediction_blocks = np.array(prediction_blocks)
+        predictions = prediction_blocks[:,0,:]
+        indices = prediction_blocks[:,1,:][:,0].astype(int) - 1
+        pred_matrix[indices] = predictions
+        return pred_matrix
```
tgsmith61591, line 331: Would you mind adding a test case for this? Maybe assert, e.g.
* Expected shape
* Expected test size
* No nans in test forecasts

tuomijal: Sure!

tgsmith61591: btw if you're on a POSIX machine (not a windows), you can use the `Makefile` for some handy targets for local testing:
* `make test-lint` will run flake8
* `make test-unit` will run the unit tests

tgsmith61591:
Not sure if you're still working on this @tuomijal but you might want to rebase on upstream master. We've made some changes that will impact the files you're working on

tuomijal:
Thank you for the notion and sorry for the delay. I have justified my laziness with the issues discussed in #364 but it seems I'm running out of excuses?

PR #372, opened 2020-08-02
[RFC] exogenous -> X

tgsmith61591:
This is an RFC proposing we switch the use of the `exogenous` arg to `X` for consistency with similar packages, and easy user onboarding. The reasons and approach are laid out in the `.rst` file in this PR.

Please leave comments, and reactions to this PR (👍  / 👎 ). I won't merge it for at least a week, and only if we decide to move forward with it.

(Inviting comments from some previous contributors @tuomijal @skyetim @mohitmunjal @christopher-siewert)

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -9,8 +9,8 @@ pmdarima: ARIMA estimators for Python
 
 ``pmdarima`` brings R's beloved ``auto.arima`` to Python, making an even stronger
 case for why you don't need R for data science. ``pmdarima`` is 100% Python + Cython
-and does not leverage any R code, and implements a single, easy-to-use
-scikit-learn-esque estimator.
+and does not leverage any R code, but is implemented in a powerful, yet easy-to-use set of
+functions & classes that will be familiar to scikit-learn users.
 
 .. toctree::
    :maxdepth: 2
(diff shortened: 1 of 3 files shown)


skyetim:
It's a great idea to change exogenous to X. One thing to further consider is that in sklearn-like estimators, it's usually
```
def fit(self, X, y=None, **kwargs)
```
I know it's a huge API change but it's always something that stops me from using a wider range of functionality in sklearn.

christopher-siewert:
I like the change to `X`.

Another option is to change it to just `exog` to match the statsmodels api. This too removes the typo issue. 

Overall `X` is better for the strategic decision of appealing to sklearn users.

tuomijal:
Sorry for delayed response. I prefer `X` over `exogenous` and `exog`. Reasons:
- It would provide symmetry with `y` that is currently used for the target. I think it would make the code very concise if we could write e.g. ```auto_arima(y=y, X=X)``` 
- I am afraid that `exog` is not widely used enough to be intuitive. Moreover, currently I keep forgetting whether the keyword is `exog` or `exogenous`. Switching to `exog` would only increase my confusion.

tgsmith61591:
Cool, I'm glad this has good positive feedback. I will make this change and starting in 1.8.0, the signature will be `(y, X, ...)`. `exogenous` will be allowed (but will generate a warning) for several release cycles afterwards.

PR #377, opened 2020-08-29
[MRG+1] Update nightly build script

aaronreidsmith:
# Description

This PR adds 2 changes:

1. A link to the failed run is added to the Slack message we receive
2. If the run fails, a table of the latest dependency versions/release dates is also sent to our Slack channel

## Type of change

# How Has This Been Tested?

- [X] Messages post to slack

# Checklist:


aaronreidsmith:
This will fail until #378 goes through. Additionally, the `Build against latest dependencies` job will fail regardless

.github/utils/get_dependency_releases.py
```diff
@@ -0,0 +1,29 @@
...
+    print(f'Usage: python {sys.argv[0]} <list of packages>')
+    sys.exit(1)
+
+packages = sys.argv[1:]
+session = requests.Session()
+
+releases = []
+for package in packages:
+    pypi = session.get(f'https://pypi.org/pypi/{package}/json').json()
+    latest_version = pypi['info']['version']
```
tgsmith61591, line 16: If this is not a 200, we could get a mystifying KeyError here. I don't think we need to handle it super well, since it's a utility, but just a heads up.

aaronreidsmith: Good call. I will have it raise before we dig into the dict

aaronreidsmith: Fixed

PR #388, opened 2020-09-09
[MRG+1] Fix upper bound in RollingForecastCV

aaronreidsmith:
# Description

This PR fixes the upper bound in our `range` as called out in #364. Should unblock #369

## Type of change

# How Has This Been Tested?

- [x] Unit tests

# Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/pmdarima/model_selection/_split.py
+++ b/pmdarima/model_selection/_split.py
@@ -242,10 +242,15 @@ class RollingForecastCV(BaseTSCrossValidator):
         # Determine the number of iterations that will take place. Must
         # guarantee that the forecasting horizon will not over-index the series
         all_indices = np.arange(n_samples)
-        for train_step_size in range(0, n_samples - h - initial + 1, step):
-            train_size = initial + train_step_size
-            train_indices = all_indices[:train_size]
-            test_indices = all_indices[train_size: train_size + h]
+        window_start = 0
+        while True:
+            window_end = window_start + initial
+            if window_end + h > n_samples:
+                break
+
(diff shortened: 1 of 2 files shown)


pmdarima/model_selection/_split.py
```diff
@@ -242,7 +242,7 @@ def _iter_train_test_indices(self, y, X):
         # Determine the number of iterations that will take place. Must
         # guarantee that the forecasting horizon will not over-index the series
         all_indices = np.arange(n_samples)
-        for train_step_size in range(0, n_samples - h - initial, step):
+        for train_step_size in range(0, n_samples - h - initial + 1, step):
```
tgsmith61591, line 245: I am not entirely convinced it's this simple. Can you add some tests like [this](https://github.com/alkaline-ml/pmdarima/blob/master/pmdarima/model_selection/tests/test_split.py#L107)?

tgsmith61591: Thinking we may have to do something more like https://github.com/alkaline-ml/pmdarima/pull/388/files#diff-583c6f60a96024f91b533cdf7a5c041eR341

aaronreidsmith: Sure let me look into that

aaronreidsmith: Added in 61983da. Basically just coped what was there. You might confirm that the test I used is actually testing what we want; I have little domain knowledge here

tgsmith61591 (approved):

PR #406, opened 2020-12-22
[MRG+2] Add fix for issue#370 to allow iterables

### LINKED ISSUE #370: seasonal_order does not accept iterable PDQ parameters
**Describe the bug**
Cannot pass an iterable to kwarg=season_order in pm.ARIMA

**To Reproduce**  

```python
import pmarima as pm
m = pm.ARIMA(order=(2, 1, 2), seasonal_order=([1, 2, 3, 52], 0, 1, 7))
m.fit(range(2000))
```

**Versions**
System:
    python: 3.7.6 (default, Dec 30 2019, 19:38:28)  [Clang 11.0.0 (clang-1100.0.33.16)]
executable: /usr/local/opt/python/bin/python3.7
   machine: Darwin-19.6.0-x86_64-i386-64bit

Python dependencies:
        pip: 20.2
 setuptools: 42.0.2
    sklearn: 0.23.1
statsmodels: 0.11.1
      numpy: 1.18.1
      scipy: 1.4.1
     Cython: 0.29.17
     pandas: 0.25.3
     joblib: 0.14.1
   pmdarima: 1.6.1

**Expected behavior**
This should create a seasonal order with lags of 1,2,3,52 but it is failing on validation to check whether the model parameters include a seasonal order

**Actual behavior**
/usr/local/lib/python3.7/site-packages/pmdarima/compat/statsmodels.py in check_seasonal_order(order)
     46         The existing seasonal order
     47     """
---> 48     if sum(order[:3]) == 0 and order[-1] == 1:
     49         order = (0, 0, 0, 0)
     50 

TypeError: unsupported operand type(s) for +: 'int' and 'list'

**Additional context**
Works fine when using iterables in the order kwarg but not in the seasonal_order kwarg

charlesdrotar:
# Description

This PR addresses the compatibility check not allowing iterables in position 0 of the the seasonal_order tuple as pointed out in #370. It first assumes an iterable at position 0 (trying to be pythonic :)), and if it iterable we return unmodified. We only override the order in the special case of `(0,0,0,1)`. Additional changes were made to add test cases to check the special case of the override and the iterable not raising a `TypeError`.

Fixes #370 

## Type of change

- [x] Bug fix (non-breaking change which fixes [issue#370](https://github.com/alkaline-ml/pmdarima/issues/370))

# How Has This Been Tested?

for your test configuration

- [x] Add test for `pmdarima.compat.statsmodels.check_seasonal_order`

# Checklist:

- [`N/A`] I have made corresponding changes to the documentation

### DIFF
--- a/pmdarima/compat/statsmodels.py
+++ b/pmdarima/compat/statsmodels.py
@@ -2,6 +2,7 @@
 #
 # Handle inconsistencies in the statsmodels API versions
 
+from collections import abc
 from pkg_resources import parse_version
 import statsmodels as sm
 
@@ -45,8 +46,18 @@ def check_seasonal_order(order):
     order : tuple
         The existing seasonal order
     """
-    if sum(order[:3]) == 0 and order[-1] == 1:
-        order = (0, 0, 0, 0)
+
+    try:
+        # Assume an iterable for order[0].
+        # If order[0] is then we don't perform check.
+        # issue#370: https://github.com/alkaline-ml/pmdarima/issues/370
+        for _ in order[0]:
+            pass
+        return order
+    except TypeError:
+        # If order[0] is not iterable we perform the check
+        if sum(order[:3]) == 0 and order[-1] == 1:
+            order = (0, 0, 0, 0)
 
     # user's order may still be invalid, but we'll let statsmodels' validation
     # handle that.
--- a/pmdarima/compat/statsmodels.py
+++ b/pmdarima/compat/statsmodels.py
@@ -2,7 +2,6 @@
 #
 # Handle inconsistencies in the statsmodels API versions
 
-from collections import abc
 from pkg_resources import parse_version
 import statsmodels as sm
 
--- a/pmdarima/compat/statsmodels.py
+++ b/pmdarima/compat/statsmodels.py
@@ -2,6 +2,7 @@
 #
 # Handle inconsistencies in the statsmodels API versions
 
+from collections.abc import Iterable
 from pkg_resources import parse_version
 import statsmodels as sm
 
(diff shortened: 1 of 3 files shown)


pmdarima/compat/statsmodels.py
```diff
@@ -45,8 +45,18 @@ def check_seasonal_order(order):
...
         The existing seasonal order
     """
-    if sum(order[:3]) == 0 and order[-1] == 1:
-        order = (0, 0, 0, 0)
+
+    try:
+        # Assume an iterable for order[0].
+        # If order[0] is then we don't perform check.
+        # issue#370: https://github.com/alkaline-ml/pmdarima/issues/370
+        for _ in order[0]:
```
aaronreidsmith, line 53: Is this an internal-only function? I ask because if not, I am not sure I am on board with this. Since a string is iterable in Python, if `order[0]` is a string, this would pass, right? Would it not be better to have something like:

```python
from collections.abc import Iterable

if isinstance(order[0], Iterable) and not isinstance(order[0], str):
    return order
else:
    if sum(order[:3]) == 0 and order[-1] == 1:
            order = (0, 0, 0, 0)
    
    return order
```

There is probably a more idiomatic way to write that, lol. On the other hand, I think this is just a pre-validation check, so if the user did pass something bad in, it would be caught by statsmodels

charlesdrotar: Ahhh good point about catching string iterables.

charlesdrotar: Done!

aaronreidsmith (approved):
LGTM!

tgsmith61591 (approved):

Issue #434, opened 2021-05-28
Add support to release aarch64 wheels


odidev:
Problem
---------
On aarch64, ‘pip install pmdarima’ builds the wheels from source code and then installs it. It requires the user to have a development environment installed on his system. Also, it takes more time to build the wheels than downloading and extracting the wheels from pypi.

Resolution
-----------
On aarch64, ‘pip install pmdarima’ should download the wheels from pypi

@aaronreidsmith, Please let me know your interest in releasing aarch64 wheels. I can help with this.

Issue #423, opened 2021-04-19
[BUG] Cant import pmdarima v1.8.1


aiwalter:
**Describe the bug**
Import error.
```
>>> import pmdarima
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\bla\blaPython\Miniconda3\envs\playground\lib\site-packages\pmdarima\__init__.py", line 51, in <module>
    from .arima import auto_arima, ARIMA, AutoARIMA, StepwiseContext, decompose
  File "C:\Users\bla\blaPython\Miniconda3\envs\playground\lib\site-packages\pmdarima\arima\__init__.py", line 5, in <module>
    from .approx import *
  File "C:\Users\bla\blaPython\Miniconda3\envs\playground\lib\site-packages\pmdarima\arima\approx.py", line 9, in <module>
    from ..utils.array import c, check_endog
  File "C:\Users\bla\blaPython\Miniconda3\envs\playground\lib\site-packages\pmdarima\utils\__init__.py", line 5, in <module>
    from .array import *
  File "C:\Users\bla\blaPython\Miniconda3\envs\playground\lib\site-packages\pmdarima\utils\array.py", line 13, in <module>
    from ._array import C_intgrt_vec
  File "pmdarima\utils\_array.pyx", line 1, in init pmdarima.utils._array
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```
**To Reproduce**  
```python
import pmdarima
```

**Versions**
Python 3.7.9.
Windows 10.
I had to reinstall `pmdarima==1.8.0` to get the dependencies:
```
        pip: 20.3.3
 setuptools: 47.3.1
    sklearn: 0.24.0
statsmodels: 0.12.2
      numpy: 1.19.4
      scipy: 1.6.0
     Cython: 0.29.17
     pandas: 1.2.4
     joblib: 1.0.0
   pmdarima: 1.8.0
```
___

Martin Walter <martin_friedrich.walter@daimler.com>, Mercedes-Benz AG on behalf of Daimler TSS GmbH.
https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md

ngupta23:
We had the same issue in `pycaret`. We had to pin it to 1.8.0 as well (previously it was >= 1.8.0).

aiwalter:
[...]

PR #435, opened 2021-05-28
[MRG+1] Add aarch64 wheel build support

odidev:
Added aarch64 wheel build support.
Related to https://github.com/alkaline-ml/pmdarima/issues/434, @aaronreidsmith Could you please review this PR?

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -141,3 +141,54 @@ jobs:
         env:
           TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
+
+  build-and-deploy-aarch64:
+    name: "Build and Deploy (aarch64 ${{ matrix.pyver }})"
+    strategy:
+      matrix:
+        pyver: [cp37-cp37m, cp38-cp38, cp39-cp39, cp310-cp310]
+        arch: [aarch64]
+      fail-fast: false
+    runs-on: ubuntu-latest
+    env:
+      py: /opt/python/${{ matrix.pyver }}/bin/python
+      img: quay.io/pypa/manylinux2014_${{ matrix.arch }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up QEMU
+      id: qemu
+      uses: docker/setup-qemu-action@v1
+    - name: Build Wheel
+      run: |
+            docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+            ${{ env.img }} \
+            bash -exc '${{ env.py }} -m pip install virtualenv && ${{ env.py }} -m venv .env && \
+            source .env/bin/activate && \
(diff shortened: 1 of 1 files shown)


aaronreidsmith:
Thanks for opening #435 and this PR! I think we'll want to do a little more testing with this architecture before approving/merging this. Additionally, just to maximize our parallelism, GitHub Actions may not be the place for it to run.

This PR also reverts our numpy changes that were necessary to support [sktime](https://github.com/alan-turing-institute/sktime) (see #423, #425, https://github.com/alan-turing-institute/sktime/pull/828, and https://github.com/alan-turing-institute/sktime/pull/831 for context), so we will need to test with our supported version of numpy. Tagging @tgsmith61591 as well to move the conversation forward.

odidev:
@tgsmith61591, Could you please have a look at this PR? Thanks in advance.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 joblib>=0.11
 Cython>=0.29,!=0.29.18
-numpy~=1.19.0
+numpy>=1.19.0
```
tgsmith61591, line 3: Same here

tgsmith61591 (approved):
Hey thanks so much for the PR @odidev. Sorry it took me a while to get to&mdash;I have a newborn at home and have been taking some time off. Could you please address the minor numpy versioning remark?

@aaronreidsmith when this passes, let's go ahead and merge up, then cut RCs from master just to make sure all wheels are distributed as expected.

odidev:
Done.

requirements.txt
aaronreidsmith: @odidev looks like this file didn't get reverted in your most recent change. Once you revert, I can merge

odidev: Done.

aaronreidsmith:
Hey @odidev, sorry to give you the run around for so long on this PR. It looks like our CI keeps timing out trying to build statsmodels from scratch for aarch64. This seems to be a [known issue that they are working on](https://github.com/MacPython/statsmodels-wheels/pull/54), but until they provide pre-built wheels, I don't think we will be able to support this. Even if we did support this, users would have to build statsmodels from scratch, which is very slow. So we'll leave this PR with a `hold` label until statsmodels supports it, and we will follow that PR closely

AGSaidi:
Looks like statsmodel has managed to release aarch64 wheels: https://pypi.org/project/statsmodels/#files 

@aaronreidsmith @odidev  perhaps we can give this another try now?

aaronreidsmith:
Nice! @odidev if you want to rebase, we can use this branch. Or I have #458 open (which was using a conda-based approach) that we can reprupose

AGSaidi:
stats model has, as expected, not built Python 3.6 wheels, so in addition to re-basing, we should drop Python 3.6 from the PR and add 3.10.

odidev:
Done re-basing. Also removed python 3.6 and added python 3.10.

aaronreidsmith:
Superseded by #458. Pretty much the same, but we changed how we test on aarch64 platforms, since that was the slow step.

PR #439, opened 2021-06-15
[MRG] Consistent prediction results for simple_differencing = True and False

### LINKED ISSUE #421: SARIMA prediction returns differenced time series when simple_differencing = True
Hi, thank you for the useful library in Python.
I'm working on a long time series with n ~ 2000 with period ~ 300.
I use auto_arima() for SARIMA with simple_differencing = True. Otherwise computation takes forever.

**Describe the bug**
With simple_differencing, predict() function returns a differenced time series.
Then I have to use diff_inv() for point forecast outside pmdarima.
Furthermore, it is not very straightforward to undifference confidence interval.
I'm wondering if pmdarima could return the undifferenced forecast and confidence interval.

**To Reproduce**  

```python
import pmdarima as pm
wineind = pm.datasets.load_wineind()

def do_fit(simple_differencing):
    return pm.auto_arima(wineind, start_p=1, start_q=1, max_p=1,
                         max_q=2, m=12, start_P=0,
                         seasonal=True,
                         d=1, D=1, stepwise=True,
                         error_action='ignore',
                         sarimax_kwargs={
                             'simple_differencing': simple_differencing
                         },
                         maxiter=2,
                         trace=True)

# simple_differencing = False
seasonal_fit = do_fit(False)
prediction = seasonal_fit.predict(n_periods=60, return_conf_int=True)
pred_mid = prediction[0]
pred_conf_int = prediction[1]

# simple_differencing = True
simple = do_fit(True)
prediction_simple = simple.predict(n_periods=60, return_conf_int=True)
pred_simple_mid = prediction_simple[0]
pred_simple_conf_int = prediction_simple[1]
```
The plot is
image

Differenced time series is returned because it inherits SARIMAX in statsmodels:
[https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.sarimax.SARIMAX.html](https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.sarimax.SARIMAX.html)

> If simple_differencing = True is used, then the endog and exog data are differenced prior to putting the model in state-space form. This has the same effect as if the user differenced the data prior to constructing the model, which has implications for using the results:

> - _Forecasts and predictions will be about the differenced data, not about the original data. (while if simple_differencing = False is used, then forecasts and predictions will be about the original data)._

**Versions**

python: 3.7.6
machine: Windows-10-10.0.18362-SP0
statsmodels: 0.11.0
pmdarima: 1.8.0

**Expected behavior**
predict() function returns undifferenced time series. This is more useful than a differenced series. Also, this behavior is more consistent with auto.arima() in R.  The R function returns undifferenced series for both approximation = true/false.

**Solution**
My change is [https://github.com/drxyzw/pmdarima/commit/4853b6267892f6595250200dddf09893c3a42ee5](https://github.com/drxyzw/pmdarima/commit/4853b6267892f6595250200dddf09893c3a42ee5)
- Adding a routine inside _seasonal_prediction_with_confidence() in arima.py.
- For forecast mid, use diff_inv() to undifference the forecast.
- Confidence interval is computed with coefficients of SARIMA and sigma2. I derive an equivalent MA model from SARIMA. With the coefficients of the equivalent MA and sigma2, we can compute variance and confidence interval.

The result is
image
Indeed, results of simple_differencing = true/false are different because the selected SARIMA orders are (0,1,2)x(0,1,1)[12] / (1,1,2)x(0,1,1)[12] respectively.

If it is fine for you, I will make a pull request from this change after cleaning tests.

drxyzw:
# Description

SARIMA and ARIMA predict() and predict_in_sample() now return undifferenced prediction and confidence intervals if simple_differencing = True.
When d or D is not 0, the two functions used to return differenced results. This was not consistent with behaviors of simple_differencing = False and corresponding R library.

Fixes #421

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Added two tests to pmdarima\pmdarima\arima\tests\test_auto.py.
The tests to check if the prediction results are similar for simple_differencing = True and False.

- [x] test_stepwise_with_simple_differencing2()
- [x] test_stepwise_with_simple_differencing3()

# Checklist:


.gitignore
```diff
@@ -1,131 +0,0 @@
-# VS Code
```
aaronreidsmith, line 1: I will let @tgsmith61591 comment on the actual implementation, but the `.gitignore` file will need to be added back before merging. It allows us to [specify files that we explicitly don't want to track](https://git-scm.com/docs/gitignore).

drxyzw: It's my bad. I will undo the removal and add .gitignore.

tgsmith61591:
@drxyzw this is a great PR, thanks so much for your contribution, and sorry for the time it took me to get back to you. I've been taking some time off.

tgsmith61591 (approved):

drxyzw:
No problem, and thank you for your review and approval.

Issue #440, opened 2021-06-30
tsdisplay errs with input arrays y of length less than 51.  


jasmyace:
**Describe the bug**
Function `tsdisplay`, via its `lag_max` default value of `50`, assumes that array `y` has at least 51 observations.  A total of (at least) 51 observations is required to produce 50 lags.  

Not directly specifying `lag_max`, when `y` is not of sufficient length, leads to a (perhaps) cryptic error.  

**To Reproduce**  

```python
import pmdarima as pm
from pmdarima.utils import tsdisplay

x = pm.datasets.load_sunspots()
x51 = x[0:51]   # 51 obs.
x50 = x[0:50]   # 50 obs. 
x49 = x[0:49]   # 49 obs.
tsdisplay(x51)  # yay!  :)
tsdisplay(x50)  # oh no!  :(
tsdisplay(x49)  # oh no!  :(
```

**Versions**
```python
System:
    python: 3.8.10 (default, May 19 2021, 11:01:55)  [Clang 10.0.0 ]
executable: /Users/Jason/opt/anaconda3/envs/python38_2/bin/python
   machine: macOS-10.14.6-x86_64-i386-64bit

Python dependencies:
        pip: 21.1.2
 setuptools: 52.0.0.post20210125
    sklearn: 0.24.2
statsmodels: 0.12.2
      numpy: 1.19.0
      scipy: 1.6.2
     Cython: 0.29.17
     pandas: 1.2.5
     joblib: 1.0.1
   pmdarima: 1.8.0
```
Not using the most recent version of `pmdarima`, but GitHub suggests most of `pmdarima.utils` has not undergone revisions recently.  

**Expected behavior**
In the case that `y` is of length 50 or less, `tsdisplay` would ideally adapt to feed `len(y) - 1`, or an appropriate equivalent, into the internal call to `plot_acf`.  For what it's worth, it looks like [`tsaplots`](https://github.com/statsmodels/statsmodels/blob/main/statsmodels/graphics/tsaplots.py) uses `np.arange` to help in this area in`plot_acf`.  

**Actual behavior**
Calling `tsdisplay` via 
```python 
tsdisplay(x50)  # oh no!  :(
```
leads to 
```python 
blahblahblah
ValueError: could not broadcast input array from shape (50) into shape (51)
```

**Additional context**
n/a

tgsmith61591:
[...]

PR #442, opened 2021-07-21
[MRG] CI consolidation + minor bug fixes

tgsmith61591:
# Description

* Circle CI changes:
  - Refactor Circle CI to build wheels during each unit test stage. This is in direct response to several issues we've had during the last several deployments, which were not discovered until the release was cut. Each unit test stage builds a wheel, runs the tests, and then persists the wheel to the workspace for the deployment stage. After merging, we will need to cut a new test release to ensure we are properly restoring the workspace.
  - Consolidate smaller, miscellaneous tests into a single stage to reduce the concurrency contention imposed by Circle CI's rate limits
* Address issue #440
* Speed up some tests & examples
* Add to documentation

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Documentation change

# How Has This Been Tested?

Everything passes (still)

# Checklist:


pmdarima/utils/visualization.py
```diff
@@ -371,6 +372,15 @@ def tsdisplay(y, lag_max=50, figsize=(8, 6), title=None, bins=25,
     ax1 = fig.add_subplot(gs[2:, 0])
     ax2 = fig.add_subplot(gs[2:, 1])
 
+    # make sure y is a np array
+    y = check_endog(y, copy=False)
+
+    if lag_max >= y.shape[0]:
+        raise ValueError(
+            f"lag_max ({lag_max}) must be < than length of the "
```
aaronreidsmith, line 380: Nit: `<` means "less than," so this reads as "lag_max (x) must be less than *than* length...". I think we should remove one or the other. I would do "less than" (with no symbol), but up to you

aaronreidsmith (approved):
LGTM. Not changing the title since you still have it marked as WIP, but looks good (with the exception of one nit) and seems to run faster!

pmdarima/utils/visualization.py
tgsmith61591: If only there were an eyeroll emoji. 😉

aaronreidsmith: Like this? 🙄

tgsmith61591: https://media.tenor.com/images/f8a02c67648240f3eba5b3fb871e7c37/tenor.gif

aaronreidsmith: pepe reeeee

:trollface:

PR #457, opened 2021-10-04
[MRG+1] Add compatibility for statsmodels 0.13 and higher

aaronreidsmith:
# Description

This should fix the nightly build. The `unbiased` keyword and methods have been deprecated in `statsmodels` 0.13.0. A couple defaults for `acf` and `pacf` have also changed

## Type of change

- [ ] Documentation change

# How Has This Been Tested?

- [X] `test_wrapped` now works if testing with `statsmodels` 0.13.0

# Checklist:

### DIFF
--- a/pmdarima/utils/wrapped.py
+++ b/pmdarima/utils/wrapped.py
@@ -58,7 +58,7 @@ def inheritdoc(parent):
 
 
 @inheritdoc(parent=sm_acf)
-def acf(x, unbiased=None, nlags=None, qstat=False, fft=True,
+def acf(x, unbiased=None, nlags=None, qstat=False, fft=None,
         alpha=None, missing='none', adjusted=False):
     kwargs = {
         "x": x,
@@ -80,6 +80,8 @@ def acf(x, unbiased=None, nlags=None, qstat=False, fft=True,
             kwargs["adjusted"] = unbiased
         else:
             kwargs["adjusted"] = adjusted
+
+        kwargs["fft"] = fft or True
     else:
         kwargs["nlags"] = nlags or 40  # Becomes `None` in 0.13.0
         kwargs["fft"] = fft or False  # Becomes `True` in 0.13.0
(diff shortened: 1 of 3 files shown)


tgsmith61591 (approved):

tgsmith61591:
Is there anything we can do to mock this in a unit test?

aaronreidsmith:
> Is there anything we can do to mock this in a unit test?

Hmmm, let me think as far as local mocking. This _is_ being tested in our pipeline, though. Statsmodels 0.13.0 is [not distributed for Python 3.6](https://pypi.org/project/statsmodels/0.13.0/#files), so our [pipeline uses 2.12.2](https://github.com/alkaline-ml/pmdarima/pull/457/checks?check_run_id=3791811483#step:9:108)

pmdarima/utils/wrapped.py
```diff
@@ -54,13 +56,60 @@ def caller(*args, **kwargs):
...
+        "x": x,
+        "nlags": nlags,
+        "qstat": qstat,
+        "fft": fft,
+        "alpha": alpha,
+        "missing": missing,
+    }
+
+    # Handle kwarg deprecation in statsmodels 0.13.0
+    if parse_version(statsmodels.__version__) >= parse_version("0.13.0"):
```
tgsmith61591, line 71: Move this into a new function def in `pmdarima.compat.statsmodels`:

```python
def _use_sm_13():
    return parse_version(statsmodels.__version__) >= parse_version("0.13.0")
```

(name it whatever you want) and then you can mock this in a unit test.

```python
from unittest import mock

@pytest.mark.parametrize("useit", [True, False])
def test_using_sm13(useit):
    with mock.patch("pmdarima.compat.statsmodels._use_sm_13") as mock_sm:
        mock_sm.return_value = useit
        x = acf(x)
        ...
```

aaronreidsmith: Had the conversation in Slack, but putting here for posterity. This will require some thinking... An example is a Python 3.9 pipeline which installs `0.13.0` or higher. If it mocks `< 0.13.0` and uses the deprecated keywords, it will fail. In order to be able to test both deprecated and non-deprecated keywords, all builds would have to be on `0.12.X`, the last (and first, actually) version to support both the old and new keywords

PR #459, opened 2021-10-07
[MRG+1] Add builds for Python 3.10

aaronreidsmith:
# Description

This PR adds builds for Python 3.10, which was [released](https://www.python.org/downloads/release/python-3100/) on 2021-10-04. WIP because we will need to wait for dependencies to support 3.10

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation change

# How Has This Been Tested?

Hasn't yet 🤠 

# Checklist:

### DIFF
--- a/build_tools/build_requirements.txt
+++ b/build_tools/build_requirements.txt
@@ -1,4 +1,5 @@
-numpy~=1.19.3
+numpy~=1.19.3; python_version < '3.10'
+numpy~=1.21.2; python_version >= '3.10'
 scipy>=1.3.2
 cython>=0.29,!=0.29.18
 scikit-learn>=0.22
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,6 +167,15 @@ jobs:
       - checkout
       - build-test-and-persist:
           pythonversion: "3.9"
+
+  test-unit-cpython310:
+    machine:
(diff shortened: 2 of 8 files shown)


aaronreidsmith:
Need to add a warning for the 32-bit deprecation, but otherwise I think this is good to go

aaronreidsmith:
> Need to add a warning for the 32-bit deprecation, but otherwise I think this is good to go

Added in 941098d

build_tools/build_requirements.txt
```diff
@@ -1,4 +1,5 @@
-numpy~=1.19.3
+numpy~=1.19.3; python_version < '3.10'
+numpy~=1.21.2; python_version >= '3.10'
```
tgsmith61591, line 2: This is cool, didn't know of this syntax

tgsmith61591 (approved):
LGTM

build_tools/build_requirements.txt
aaronreidsmith: Yeah, I didn't either! In this case it is because numpy doesn't distribute for 3.10 on 1.19.X, and we need to pin the lowest supported version. It also works with things like platform and whatnot

https://pip.pypa.io/en/stable/cli/pip_install/#requirement-specifiers

PR #496, opened 2022-04-17
[MRG+1] Add 'concurrency' block to GHA and fix build warnings

aaronreidsmith:
# Description

This PR makes it so if we push a commit to the same branch, it will cancel previous builds on GHA, saving us some cycles.

It also fixes a couple warnings in the build:

- `np.int` -> `int`
- `np.bool` -> `bool`
- `np.float` -> `float`
- `LinearRegression(normalize=True)` was apparently deprecated by sklearn in favor of `make_pipeline(StandardScaler(with_mean=False), LinearRegression())`

    ```
    /Users/runner/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/sklearn/linear_model/_base.py:145: FutureWarning: 'normalize' was deprecated in version 1.0 and will be removed in 1.2.
  If you wish to scale the data, use Pipeline with a StandardScaler in a preprocessing stage. To reproduce the previous behavior:
  
  from sklearn.pipeline import make_pipeline
  
  model = make_pipeline(StandardScaler(with_mean=False), LinearRegression())
  
  If you wish to pass a sample_weight parameter, you need to pass it as a fit parameter to each step of the pipeline as follows:
  
  kwargs = {s[0] + '__sample_weight': sample_weight for s in model.steps}
  model.fit(X, y, **kwargs)
    ``` 

## Type of change

- [X] CI/CD change
- [X] Fix deprecations

# How Has This Been Tested?

- [X] Tests still pass
- [X] Most warnings are gone
- [X] Old builds cancel on push to same branch

# Checklist:

N/A

### DIFF
--- a/pmdarima/arima/seasonality.py
+++ b/pmdarima/arima/seasonality.py
@@ -8,6 +8,8 @@ from collections import namedtuple
 
 import math
 from sklearn.linear_model import LinearRegression
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
 
 from scipy.linalg import svd
 from statsmodels import api as sm
@@ -226,12 +228,15 @@ class CHTest(_SeasonalStationarityTest):
 
         # no use checking, because this is an internal method
         # if n <= s:  raise ValueError('too few samples (%i<=%i)' % (n, s))
-        frec = np.ones(int((s + 1) / 2), dtype=np.int)
+        frec = np.ones(int((s + 1) / 2), dtype=int)
         ltrunc = int(np.round(s * ((n / 100.0) ** 0.25)))
         R1 = CHTest._seas_dummy(wts, s)
 
         # fit model, get residuals
-        lmch = LinearRegression(normalize=True).fit(R1, wts)
+        lmch = make_pipeline(
+            StandardScaler(with_mean=False),
+            LinearRegression()
+        ).fit(R1, wts)
         # lmch = sm.OLS(wts, R1).fit(method='qr')
         residuals = wts - lmch.predict(R1)
 
--- a/pmdarima/utils/tests/test_array.py
+++ b/pmdarima/utils/tests/test_array.py
@@ -135,7 +135,7 @@ def test_diff():
 )
 def test_diff_inv(arr, lag, differences, xi, expected):
     res = diff_inv(arr, lag=lag, differences=differences, xi=xi)
-    expected = np.array(expected, dtype=np.float)
+    expected = np.array(expected, dtype=float)
     assert_array_equal(expected, res)
 
 
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -11,6 +11,11 @@ on:
     branches:
       - '*'
 
+# Cancel older runs of the same workflow on the same branch
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build-and-deploy:
     name: Build and Deploy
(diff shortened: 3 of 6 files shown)


aaronreidsmith:
Only remaining warnings are these, which I think is fine? They are on our side at least, and I think we are intentionally creating those warnings. Could be wrong

```
=============================== warnings summary ===============================
tests/test_pipeline.py::test_pipeline_predict_inverse_transform[True-True-X0-pipeline5]
tests/test_pipeline.py::test_pipeline_predict_inverse_transform[True-True-X1-pipeline5]
  /Users/runner/hostedtoolcache/Python/3.7.12/x[64](https://github.com/alkaline-ml/pmdarima/runs/6054174506?check_suite_focus=true#step:11:64)/lib/python3.7/site-packages/pmdarima/preprocessing/endog/boxcox.py:161: RuntimeWarning: overflow encountered in exp
    return np.exp(y) - lam2, exog

tests/test_pipeline.py::test_pipeline_predict_inverse_transform[True-True-X0-pipeline6]
tests/test_pipeline.py::test_pipeline_predict_inverse_transform[True-True-X1-pipeline6]
  /Users/runner/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pmdarima/preprocessing/endog/boxcox.py:165: RuntimeWarning: invalid value encountered in power
    de_exp = numer ** (1. / lam1)  # de-exponentiate
```

pmdarima/arima/seasonality.py
```diff
@@ -226,12 +228,15 @@ def _sd_test(wts, s):
...
+        frec = np.ones(int((s + 1) / 2), dtype=int)
         ltrunc = int(np.round(s * ((n / 100.0) ** 0.25)))
         R1 = CHTest._seas_dummy(wts, s)
 
         # fit model, get residuals
-        lmch = LinearRegression(normalize=True).fit(R1, wts)
+        lmch = make_pipeline(
+            StandardScaler(with_mean=False),
+            LinearRegression()
+        ).fit(R1, wts)
```
tgsmith61591, line 239: I can't really remember everything that leads up to this call, but are we assuming data is already centered? (e.g., `with_mean=False`)

pmdarima/utils/tests/test_array.py
```diff
@@ -135,7 +135,7 @@ def test_diff():
 )
 def test_diff_inv(arr, lag, differences, xi, expected):
     res = diff_inv(arr, lag=lag, differences=differences, xi=xi)
-    expected = np.array(expected, dtype=np.float)
+    expected = np.array(expected, dtype=float)
```
tgsmith61591, line 138: OOC what is the impetus for these dtype changes?

tgsmith61591 (approved):
🔥

pmdarima/utils/tests/test_array.py
aaronreidsmith: The impetus on our side were the build warnings like this:

```
DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    frec = np.ones(int((s + 1) / 2), dtype=np.int)
```

On the [numpy side](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations) "naked" dtypes seem to have been aliases for a while and just because deprecated. So to get rid of the warning the options were to just use the Python built-in or to add a precision (with `np.bool` only having one option, obviously).

Basically I am just trying to stay on top of warnings in case things ever get removed. That has bitten us in the past 🤷🏻

PR #498, opened 2022-04-18
[MRG+1] Sunset 32-bit builds

aaronreidsmith:
# Description

Windows is 1-2% of our user base ([source](https://pypistats.org/packages/pmdarima)) with 32-bit probably being a small fraction of that. Microsoft stopped offering 32-bit Windows in 2020, so there is no need to continue supporting it. We already stopped building 32-bit for 3.10+ due to dependencies not supporting it, v2.0 seems as good a time as ever to remove it completely. This should also help our build times since it reclaims 3 builds on each commit in GHA.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] CI/CD change

# How Has This Been Tested?

N/A

# Checklist:

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -23,16 +23,6 @@ jobs:
       matrix:
         os: [windows-latest, macos-10.15]
         python-version: ['3.7', '3.8', '3.9', '3.10']
-        architecture: ['x86', 'x64']
-        exclude:
-          # Don't build 32-bit on Mac
-          - os: macos-10.15
-            architecture: x86
-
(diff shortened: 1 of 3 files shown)


aaronreidsmith:
Had to update the "require checks" to match the new builds, so this will block other PRs for now

.github/workflows/build_and_deploy.yml
```diff
@@ -49,7 +39,6 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-          architecture: ${{ matrix.architecture }}
```
aaronreidsmith, line 52: This defaults to `x64`, which is why it was removed

tgsmith61591 (approved):

PR #547, opened 2023-05-08
[MRG] Migrate to pyproject.toml

aaronreidsmith:
# Description

[`pip` 23.1](https://pip.pypa.io/en/stable/news/#v23-1) drops support for `setup.py` being called directly (`python setup.py`), so we have to adapt by adding a `pyproject.toml` as an entrypoint. I am using `setuptools.build_meta:__legacy__` as the build backend because it is the most similar to what we have today. Eventually we'll want to move to something else `setuptools.build_meta`, `meson`, etc.). I ran into problems trying to do those directly because they don't allow partial imports of `pmdarima` during the build process.

Separately, `distutils` (and by extension `numpy.distutils`) is deprecated and being removed in Pytohn 3.12+. We need to look into how to replace that... 

## Type of change

- [X] Build change

# How Has This Been Tested?

- [X] Builds pass

# Checklist:


aaronreidsmith:
> Separately, distutils (and by extension numpy.distutils) is deprecated and being removed in Pytohn 3.12+. We need to look into how to replace that...

Looks like we can use `setuputils.command.build_ext` as a drop-in replacement. I might try that on this branch

tgsmith61591 (approved):

